PR #8, opened 2024-05-17
Don't delete `Class.props` on adding props, closes #6

### LINKED ISSUE #6: Adding props should never delete `Class.props`
We recently had a bug where because `defineEvents()` was called before prop initialization, the props it was defining were *overwriting* `Class.props`.

This is a huge footgun and should never happen.

I can see a few solutions.
So, when `Class.props` is already a `Props` object, adding props will just add them to that `Props` object. So the problem only happens when `Class.props` exists but is *not* (yet?) a `Props` object.

1. We initialize it then. Cons: Obtrusive. What if the user is trying to combine a mixin from Nude Element with some other library to handle props?
2. We store it for future reference. Instead of `props` defaulting to just `Class.props`, it would default to `Class[propsSymbol] ?? Class.props` where `Class[propsSymbol]` would be a props object we have found earlier on that class. I think this is the lease obtrusive solution, but still maintains the ergonomics of 1.

DmitrySharabin:
Instead, store it in `Class[propsSymbol]` for future reference.

### DIFF
--- a/src/props/defineProps.js
+++ b/src/props/defineProps.js
@@ -5,12 +5,13 @@ import {
 	queueInitFunction,
 } from "../util.js";
 
+let propsSymbol = Symbol("propsSymbol");
 
 function init () {
 	this.constructor.props.initializeFor(this);
 }
 
-export default function defineProps (Class, props = Class.props) {
+export default function defineProps (Class, props = Class[propsSymbol] ?? Class.props) {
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
To test that it works as expected, I placed `defineEvents()` before `defineProps()` (as it was once) in the `NudeElement` class, and it didn't break anything.

LeaVerou (approved):

PR #9, opened 2024-05-17
Default values should be parsed, closes #7

### LINKED ISSUE #7: `parse()` should be invoked on a prop's default value

DmitrySharabin:
It attempts to fix #7 by invoking `parse()`, if present, on a prop's resolved default value.

At the same time, it fixes the https://github.com/color-js/elements/issues/39 issue with `<color-picker>` when `space` is not provided.

### DIFF
--- a/src/props/Prop.js
+++ b/src/props/Prop.js
@@ -137,7 +137,8 @@ let Self = class Prop {
 					return this.defaultProp.get(element);
 				}
 				else {
-					return resolveValue(this.default, [element, element]);
+					value = resolveValue(this.default, [element, element]);
+					return this.parse?.(value) ?? value;
 				}
 			}
 		}


src/props/Prop.js
```diff
@@ -137,7 +137,8 @@ let Self = class Prop {
 					return this.defaultProp.get(element);
 				}
 				else {
-					return resolveValue(this.default, [element, element]);
+					value = resolveValue(this.default, [element, element]);
+					return this.parse?.(value) ?? value;
```
LeaVerou, line 141: Nit: is there any chance `this.parse` will not be defined? I thought there’s always one there, even if it just does the default stuff.

LeaVerou (approved):
LGTM, see nit but it's not a blocker

src/props/Prop.js
DmitrySharabin: I also thought about it, and that's why I ended up with optional chaining here: `this.parse?.(value)`. I also checked in the console; everything should work fine.

image

Is there something I'm missing?

DmitrySharabin: 🫣 I misread your comment. Sorry. Yes, I know you know it would work.

PR #17, opened 2024-05-28
First stab at Object type

LeaVerou:
Haven't tried it, feel free to merge if it works!

### DIFF
--- a/src/props/types.js
+++ b/src/props/types.js
@@ -132,3 +132,69 @@ types.set(Array, {
 		return value.join(joiner);
 	},
 });
+
+types.set(Object, {
+	equals (a, b, { valueType } = {}) {
+		let aKeys = Object.keys(a);
+		let bKeys = Object.keys(b);
+
+		if (aKeys.length !== bKeys.length) {
+			return false;
+		}
+
+		return aKeys.every(key => equals(a[key], b[key], valueType));
+	},
+
+	/**
+	 * Parses a simple microsyntax for declaring key-value options:
+	 * If no value is provided, it becomes `true`. The string "false" is parsed as `false`.
+	 * Escapes for separators are supported, via backslash.
+	 * @param {string} value
+	 * @param {Object} [options]
+	 * @param {Function} [options.valueType] The type to parse the values as
+	 * @param {boolean} [options.useMap] Whether to return a Map instead of an object
+	 * @param {string} [options.separator=","] The separator between entries.
+	 */
+	parse (value, { valueType, useMap, separator = ", " } = {}) {
+		separator = separator.trim();
+		let entrySeparatorRegex = RegExp(`\\s*(?<!\\\\)${separator}\\s*`);
+		let entries = value.trim().split(entrySeparatorRegex);
+
+		entries = entries.map(entry => {
+			let parts = entry.split(/(?<!\\):/);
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
Is there a need to trim values?

For example, `parse("foo: ba\\,r, key: value")` returns `{foo: " ba\\,r", "key": " value"}` where both values have spaces before the actual value.

DmitrySharabin:
> Is there a need to trim values?

If we don't, `parse("key: false")` becomes `{ key: " false" }` instead of `{ key: false }`.

src/props/types.js
```diff
@@ -131,4 +131,70 @@ types.set(Array, {
...
+types.set(Object, {
+	equals (a, b, { valueType } = {}) {
+		let aKeys = Object.keys(a);
+		let bKeys = Object.keys(b);
+
+		if (aKeys.length !== bKeys.length) {
+			return false;
+		}
+
+		return aKeys.every(key => equals(a[key], b[key], valueType));
```
DmitrySharabin, line 145: Is it a recursive call? Or do we expect to use some other `equals` function inside `every()`?

DmitrySharabin: I think we do, but it seems shadowed by the `equals` function we are in.

src/props/types.js
```diff
@@ -131,4 +131,70 @@ types.set(Array, {
...
+	 * @param {boolean} [options.useMap] Whether to return a Map instead of an object
+	 * @param {string} [options.separator=","] The separator between entries.
+	 */
+	parse (value, { valueType, useMap, separator = ", " } = {}) {
+		separator = separator.trim();
+		let entrySeparatorRegex = RegExp(`\\s*(?<!\\\\)${separator}\\s*`);
+		let entries = value.trim().split(entrySeparatorRegex);
+
+		entries = entries.map(entry => {
+			let parts = entry.split(/(?<!\\):/);
```
DmitrySharabin, line 164:
```suggestion
			let parts = entry.split(/(?<!\\):\s*/);
```

This should fix issues with extra spaces.

LeaVerou:
Gah, I removed the `\s*` from the regex because then re-joining is lossy, and was planning to trim manually (by calling `trim()`) but I forgot. Can you add that?

src/props/types.js
```diff
@@ -131,4 +131,70 @@ types.set(Array, {
...
+			if (parts.length > 2) {
+				// Value contains colons
+				key = parts.shift();
+				value = parts.join(":");
+			}
+			else if (parts.length > 0) {
+				key = parts[0];
+				value = parts.length > 1 ? parts[1] : true;
+			}
+
```
DmitrySharabin, line 176:
```suggestion

			value = value.trim();
```

DmitrySharabin: Like so?

LeaVerou: That will fail if they're not strings. Just pushed a fix, lemme know if it works.

DmitrySharabin: I does work!

LeaVerou:
Does that mean it can be merged?

DmitrySharabin:
I tested it a bit. And yes, I think it can be merged. It took me some time to figure out how it works, though. :)

Issue #16, opened 2024-05-24
Support arbitrary mixins


LeaVerou:
This came out of #15 . 

I've been thinking that Nude Element could also serve as a framework for arbitrary element mixins, and mixins could return either a function (which would be an init function, just like now) or an object with various functions to run at various points in the component’s lifecycle (`init`, `connected`, `disconnected`, `adopted`, `propChanged`). That could address the problem in #15 , AND provide authors useful functionality in one fell swoop.

For mixins, authors would be responsible for executing the lifecycle functions, though the mixin can check for certain data structures and automatically add them there (as a generalization of `initQueue`). Perhaps these can be named `lifecycleQueue` with keys like `init`, `connected` etc. For NudeElement, it would take care of automatically executing them.

We already call `queueInitFunction()` in every mixin, this should require no more effort by the mixin author than that. We'd just move that function to a separate module `src/mixins/defineMixin.js` that would take care of this.

PR #18, opened 2024-05-28
First stab at mixins, rel #16

LeaVerou:


LeaVerou:
I've done some testing, fixed some bugs, so I'm merging this. @DmitrySharabin feel free to take a look when you wake up in case you find any issues!

PR #20, opened 2024-05-31
Fix typo in `parse` for Maps

DmitrySharabin:
Added the `else` block when evaluating `entries`.


DmitrySharabin:
This PR is out of date and incorrect. So, I'm gonna close it.

PR #31, opened 2024-06-10
Pass resolved type to `parse()`, fixes #27

### LINKED ISSUE #27: Can't specify `Object` as type because the built-in `Object` class already has its own properties (methods) `keys` and `values`
I stumbled on this issue when I tried to define the `coords` property like so:

```js
coords: {
	type: {
		is: Array,
		values: Object,
	},
```

When we parse a value and pass it along with its type to the `parseEntries()` method https://github.com/nudeui/element/blob/eebac9359bf0eb32da7e3fb406716f35a7247677/src/props/types/dictionaries.js#L4
we destructure the `keys` property (a function), even though we didn't specify it in our type when defining the prop. It was destructured from the built-in `Object` class.

On the next step, `keys` is defined, and the following block of code is executed

https://github.com/nudeui/element/blob/eebac9359bf0eb32da7e3fb406716f35a7247677/src/props/types/dictionaries.js#L43-L45

the generic `parse()` function is called with `keys()` as a type. And of course this function throws on this line

https://github.com/nudeui/element/blob/eebac9359bf0eb32da7e3fb406716f35a7247677/src/props/types/basic.js#L25

I can also suggest that we'll face the same issue with `Map`s.

We have a workaround, but I think everything should work without it:

```js
coords: {
	type: {
		is: Array,
		values: { is: Object },
	},
```

DmitrySharabin:


LeaVerou (changes requested):
Don’t call `resolve()` twice, just set `type = `resolve(type)` and then destructure that

DmitrySharabin:
> Don’t call `resolve()` twice, just set `type = `resolve(type)` and then destructure that

Done!

LeaVerou (approved):

PR #33, opened 2024-06-18
rawProp start

LeaVerou:


src/props/Prop.js
```diff
@@ -55,6 +62,10 @@ let Self = class Prop {
 				...inferDependencies(spec.convert),
 				...(spec.additionalDependencies ?? []),
 			]);
+
+			if (this.rawProp) {
```
DmitrySharabin, line 66: I believe here should be `spec.rawProp`?

src/props/Prop.js
```diff
@@ -55,6 +62,10 @@ let Self = class Prop {
...
+				this.dependencies.add(this.rawProp);
```
DmitrySharabin, line 67: And here?

src/props/Props.js
```diff
@@ -120,6 +120,13 @@ export default class Props extends Map {
 			}
 		}
 
+		if (prop.rawProp && element.props[prop.name] !== undefined && element.props[prop.rawProp] !== undefined) {
+			delete element.props[prop.rawProp];
+			let rawProp = this.get(prop.rawProp);
+			rawProp.defaultProp ??= prop;
+			this.propChanged(element, rawProp, {source: "property", value: element.props[prop.rawProp]});
```
DmitrySharabin, line 127: By this time, `element.props[prop.rawProp]` is already deleted, no? What `value` will be equal to?

DmitrySharabin (changes requested):
I have a couple of questions. 😃

PR #37, opened 2024-06-20
[Prop] Add tests for defaults

DmitrySharabin:


test/Prop.js
```diff
@@ -117,6 +118,73 @@ export default {
...
+							description: "What should be used if both default and defaultProp are specified? If defaultProp, where should default go?",
+							arg: {
+								default: "foo",
+								defaultProp: "bar",
+							},
+							expect: "foo", // ??
+						},
+						{
+							name: "Function and prop",
+							description: "What should be used if both default and defaultProp are specified? If defaultProp, where should default go?",
```
LeaVerou, line 177: If we don't know, skip

test/Prop.js
```diff
@@ -117,6 +118,73 @@ export default {
...
+							arg: {
+								default: "foo",
+							},
+							expect: "foo",
+						},
+						{
+							name: "Function",
+							arg: {
+								default () {
+									return "foo";
```
LeaVerou, line 154: Might be good to use something other than the name of the prop here, for clarity. E.g. a number. Since you have other tests that return prop names, it took me some time to realize you were testing values.

LeaVerou (approved):

test/Prop.js
DmitrySharabin: I just wanted to draw your attention to this thing. :)

I will add `skip`.

PR #41, opened 2024-10-25
[Props][Prop] Delay the `propchange` events for uninitialized props

DmitrySharabin:
This helps us solve the problem with props set programmatically before elements are fully initialized.

### DIFF
--- a/src/props/Props.js
+++ b/src/props/Props.js
@@ -12,6 +12,8 @@ export default class Props extends Map {
 	 */
 	dependents = {};
 
+	#initialized = false;
+
 	/**
 	 *
 	 * @param {HTMLElement} Class The class to define props for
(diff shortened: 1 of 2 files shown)


src/props/Props.js
```diff
@@ -160,6 +162,8 @@ export default class Props extends Map {
 			prop.initializeFor(element);
 		}
 
+		this.#initialized = true;
```
LeaVerou, line 165: Hm. This means that the first time this runs it is _always_ deferred. Is there a way to only defer when actually needed?

DmitrySharabin: Good question. I need to think about it.

DmitrySharabin:
You were absolutely right: We should only defer the `propchange` event for uninitialized props, not all of them. This should be fixed now.

LeaVerou (approved):

PR #42, opened 2024-10-28
[Props] Take dependency graph into account when working with props

DmitrySharabin:
In this PR, we reorder props and their dependents according to the dependency graph. This means that when we initialize props, update dependencies, etc., we work with props, respecting their order in the dependency graph. This should fix the issues when we update props in the wrong order.

### DIFF
--- a/src/props/Props.js
+++ b/src/props/Props.js
@@ -1,5 +1,5 @@
 import {
-	sortObject,
+	sortMap,
 } from "./util.js";
 import Prop from "./Prop.js";
 import PropChangeEvent from "./PropChangeEvent.js";
@@ -7,10 +7,10 @@ import PropChangeEvent from "./PropChangeEvent.js";
 export default class Props extends Map {
 	/**
 	 * Dependency graph
-	 * @type {Object.<string, Set<string>>}
-	 * Key is the name of the prop, value is a set of prop names that depend on it
+	 * @type {Map.<string, Set<Prop>>}
+	 * Key is the name of the prop, value is a set of props that depend on it
 	 */
-	dependents = {};
+	dependents = new Map();
 
 	/**
 	 *
@@ -48,10 +48,10 @@ export default class Props extends Map {
 
 	updateDependents () {
 		// Rebuild dependency graph
-		let dependents = {};
+		let dependents = new Map();
 
 		for (let name of this.keys()) {
-			dependents[name] = new Set();
+			dependents.set(name, new Set());
 		}
 
 		let keyIndices = Object.fromEntries([...this.keys()].map((key, i) => [key, i]));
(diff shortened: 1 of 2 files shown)


src/props/Props.js
```diff
@@ -76,21 +76,47 @@ export default class Props extends Map {
 				dependents[name]?.add(prop);
 
 				// Dependent props should come after the prop they depend on
-				if (keyIndices[name] < keyIndices[prop.name]) {
+				if (keyIndices[name] > keyIndices[prop.name]) {
```
LeaVerou, line 79: Yikes, does that mean that until now we were basically sorting them in the exact opposite order?

src/props/Props.js
```diff
@@ -76,21 +76,47 @@ export default class Props extends Map {
...
+			// Reorder the props according to their order in the dependency graph
+			// so that every time we use this.values() or this.keys(), the result is in the correct order
+			for (let propName of Object.keys(this.dependents)) {
+				let prop = this.get(propName);
+				this.delete(propName);
+				this.set(propName, prop);
+			}
+
+			// Reorder dependents of every prop according to the dependency graph built
+			for (let [name, dependents] of Object.entries(this.dependents)) {
```
LeaVerou, line 107: Should this be in `Prop`? (not sure)

src/props/Props.js
```diff
@@ -76,21 +76,47 @@ export default class Props extends Map {
...
+		else {
 			// Sort dependency graph using the new order in keyIndices
 			// TODO put props with no dependencies first
 			// TODO do we need a topological sort?
 			this.dependents = sortObject(dependents, ([a], [b]) => {
 				return keyIndices[a] - keyIndices[b];
 			});
+
+			// Reorder the props according to their order in the dependency graph
+			// so that every time we use this.values() or this.keys(), the result is in the correct order
```
LeaVerou, line 99: Would it make sense to have a flag about whether the order actually changed before we embark on this?

LeaVerou (approved):
LGTM, see minor comments

src/props/Props.js
DmitrySharabin: It looks like so. It didn't cause any problems, though, since we didn't use the dependency graph much.

src/props/Props.js
DmitrySharabin: I think it would (actually, I was thinking about it, too). Let me think about it a bit more.

DmitrySharabin: Oh, the `sort` flag does precisely that. We rearrange the props in case they are in the wrong order (when `sort` equals `true` and we need to sort the `dependents` object). So, I assume we already do reordering _only_ when we need it.

PR #43, opened 2024-12-23
[Element] Improve support for global styles

DmitrySharabin:
- Accept URLs, strings with rules, and an array of either.

- Use `TAG-NAME-global-style-ready` classes as a style hook.

- Allow custom things via the `TAG-NAME-global-style-ready` hook.

### DIFF
--- a/src/Element.js
+++ b/src/Element.js
@@ -37,11 +37,19 @@ const Self = class NudeElement extends HTMLElement {
 
 		if (this.constructor.globalStyleSheet) {
 			let rootNode = this.getRootNode();
+			let stylesheet = rootNode.querySelector(`style[data-for="${this.constructor.tagName}"]`);
 
-			if (!rootNode.querySelector(`style[data-for="${this.constructor.tagName}"]`)) {
+			if (!stylesheet) {
 				let root = rootNode.nodeType === Node.DOCUMENT_NODE ? rootNode.head : rootNode;
-				root.append(this.constructor.globalStyleSheet.cloneNode(true));
+				stylesheet = this.constructor.globalStyleSheet.cloneNode(true);
+				root.append(stylesheet);
 			}
+
+			stylesheet.onload = () => {
+				let name = `${this.constructor.tagName}-global-style-ready`;
+				(rootNode.body ?? rootNode).classList.add(name);
+				this.constructor.hooks.run(name, this);
+			};
 		}
 
 		this.constructor.hooks.run("connected", this);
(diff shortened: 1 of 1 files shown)


LeaVerou:
> Accept URLs, strings with rules, and an array of either.

This should apply to shadow styles too.

src/Element.js
```diff
@@ -74,7 +82,24 @@ const Self = class NudeElement extends HTMLElement {
...
 			this.globalStyleSheet.dataset.for = this.tagName;
-			this.globalStyleSheet.textContent = `@import url("${this.globalStyle}")`;
+
+			let styles = Array.isArray(this.globalStyle) ? this.globalStyle : [this.globalStyle];
+			let imports = [];
+			let rules = [];
+			styles.forEach(style => {
+				// Try converting `style` to a URL and using it inside an `@import` rule; otherwise, use it as-is
+				try {
+					let url = new URL(style);
```
LeaVerou, line 92: I would use https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static when available. Maybe wrap it in a helper method or add a polyfill?

src/Element.js
```diff
@@ -74,7 +82,24 @@ const Self = class NudeElement extends HTMLElement {
...
+					imports.push(`@import url("${ url.href }");`);
```
LeaVerou, line 93: Why would this error? It shouldn't be in the try catch

src/Element.js
```diff
@@ -74,7 +82,24 @@ const Self = class NudeElement extends HTMLElement {
...
+				}
+				catch {
+					rules.push(style);
+				}
+
+			});
+
+			// Place `@import` rules at the top of the stylesheet, preserving the original order of the URLs
+			this.globalStyleSheet.textContent = [...imports, ...rules].join("\n");
```
LeaVerou, line 102: We should use adoptedStylesheets if using verbatim CSS code, ideally with a map so we can pick the same stylesheet for the same code.

src/Element.js
```diff
@@ -37,11 +37,19 @@ const Self = class NudeElement extends HTMLElement {
...
-			if (!rootNode.querySelector(`style[data-for="${this.constructor.tagName}"]`)) {
+			if (!stylesheet) {
 				let root = rootNode.nodeType === Node.DOCUMENT_NODE ? rootNode.head : rootNode;
-				root.append(this.constructor.globalStyleSheet.cloneNode(true));
+				stylesheet = this.constructor.globalStyleSheet.cloneNode(true);
+				root.append(stylesheet);
 			}
+
+			stylesheet.onload = () => {
+				let name = `${this.constructor.tagName}-global-style-ready`;
```
LeaVerou, line 49: I don't understand what this class is for.

LeaVerou:
> * Use `TAG-NAME-global-style-ready` classes as a style hook.
> * Allow custom things via the `TAG-NAME-global-style-ready` hook.

I don't understand what this style hook does or how it's supposed to be used.

DmitrySharabin:
> > * Use `TAG-NAME-global-style-ready` classes as a style hook.
> > * Allow custom things via the `TAG-NAME-global-style-ready` hook.
> 
> I don't understand what this style hook does or how it's supposed to be used.

I was thinking of something like this:

```css
body:not(.color-chart-global-style-ready) color-chart {
  opacity: 0;
}
```

Right now, the main issue we have with `<color-chart>` is that global styles are not fully applied even when the chart element is defined. This style hook should help (from my perspective).

LeaVerou:
> global styles are not fully applied even when the chart element is defined. This style hook should help (from my perspective).

Why?

Would it help if `customElements.define()` was not called until the stylesheet had loaded? Or if adoptedStylesheets are supported, we could `fetch()` the stylesheet and use it that way.

DmitrySharabin:
I'll close this in favor of #44 for now. We can use some ideas from this PR a bit later.

PR #44, opened 2025-01-16
[Element] Improve performance and reduce FOUC by preloading global stylesheets, not `@import` them

DmitrySharabin:
This @LeaVerou's proposal seems to work very well. We still need to eliminate the flash of undefined custom elements. But that's for a separate PR.

### DIFF
--- a/src/Element.js
+++ b/src/Element.js
@@ -35,15 +35,6 @@ const Self = class NudeElement extends HTMLElement {
 			this[instanceInitialized] = true;
 		}
 
-		if (this.constructor.globalStyleSheet) {
-			let rootNode = this.getRootNode();
-
-			if (!rootNode.querySelector(`style[data-for="${this.constructor.tagName}"]`)) {
-				let root = rootNode.nodeType === Node.DOCUMENT_NODE ? rootNode.head : rootNode;
-				root.append(this.constructor.globalStyleSheet.cloneNode(true));
-			}
-		}
-
 		this.constructor.hooks.run("connected", this);
 	}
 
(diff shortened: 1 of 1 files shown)


src/Element.js
```diff
@@ -72,9 +63,19 @@ const Self = class NudeElement extends HTMLElement {
...
 		if (this.globalStyle) {
-			this.globalStyleSheet = document.createElement("style");
-			this.globalStyleSheet.dataset.for = this.tagName;
-			this.globalStyleSheet.textContent = `@import url("${this.globalStyle}")`;
+			let link = document.createElement("link");
+
+			// Initiate the download immediately
+			link.rel = "preload";
+			link.as = "style";
+			link.href = this.globalStyle;
```
LeaVerou, line 71:
```suggestion
			// Initiate the download immediately
			let link = Object.assign(document.createElement("link"), {
				rel: "preload",
				as: "style",
				href: this.globalStyle,
			});
```

src/Element.js
```diff
@@ -72,9 +63,19 @@ const Self = class NudeElement extends HTMLElement {
...
+
+			link.onload = () => {
+				// Once loaded, apply the styles
+				link.rel = "stylesheet";
```
LeaVerou, line 75: Why not just create a regular stylesheet that will automatically apply the styles once loaded? What does that save us?

src/Element.js
```diff
@@ -72,9 +63,19 @@ const Self = class NudeElement extends HTMLElement {
...
+			};
+
+			document.head.append(link);
```
LeaVerou, line 78: Oh I see. You apply a single stylesheet to the head. 
What I was referring to is using a `<link>` element **in the shadow DOM**.

src/Element.js
DmitrySharabin: At first, I thought it would load and apply faster than a regular stylesheet, but after checking both options, I realized it was not. So, I switched to the regular stylesheet approach. Thank you!

src/Element.js
DmitrySharabin: I'll do it there, too. :)

However, switching from `@import` to `<link>` here also helps. Noticeably. Especially with `<color-chart>`.

src/Element.js
```diff
@@ -72,9 +63,12 @@ const Self = class NudeElement extends HTMLElement {
 		}
 
 		if (this.globalStyle) {
-			this.globalStyleSheet = document.createElement("style");
-			this.globalStyleSheet.dataset.for = this.tagName;
-			this.globalStyleSheet.textContent = `@import url("${this.globalStyle}")`;
+			let link = Object.assign(document.createElement("link"), {
+				rel: "stylesheet",
+				href: this.globalStyle,
```
LeaVerou, line 68: I wonder if we also want to add `blocking=render` and/or [`fetchpriority=high`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#fetchpriority)

DmitrySharabin: Oh, that might be a good idea. I'll first change the color elements to see what we have and then try this out.

PR #53, opened 2025-10-14
Move `Element` to use class mixins

LeaVerou:


src/util/adopt-css.js
```diff
@@ -12,48 +12,87 @@ const adoptedStyleSheets = new WeakMap();
...
-		if (!root.adoptedStyleSheets.includes(sheet)) {
-			// Not already adopted, so we need to adopt it
-			if (Object.isFrozen(root.adoptedStyleSheets)) {
-				// Slightly older browsers
-				root.adoptedStyleSheets = [...root.adoptedStyleSheets, sheet];
-			}
-			else {
-				root.adoptedStyleSheets.push(sheet);
-			}
+	// If we’re here, we have no existing object and create is true
```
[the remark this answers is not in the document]
LeaVerou: Ah yes, basic Boolean algebra — the opposite of `!A && !B` is not `A && B`, it's `A || B`! 🤦🏽‍♀️

src/styles/shadow.js
```diff
@@ -1,47 +1,70 @@
...
-		}
-	},
+export const Mixin = (Super = HTMLElement) => class ShadowStyles extends Super {
 	async init () {
 		if (!this.shadowRoot) {
 			return;
 		}
 
+		this.constructor.init();
+		this[render]();
```
DmitrySharabin: I wonder if we should `await` the result of `this[render]()` here. If not, there's probably no need for `async` on line 13.

src/styles/shadow.js
```diff
@@ -1,47 +1,77 @@
...
+		}
+
+		this[initialized] = true;
+
+		let supers = getSupers(this, HTMLElement);
+
+		for (let Class of supers) {
+			if (Object.hasOwn(Class, styles) && !Object.hasOwn(Class, fetchedStyles)) {
+				// Initiate fetching when the first element is constructed
+				let styles = (Class[fetchedStyles] = Array.isArray(Class[styles])
```
DmitrySharabin: Here, we have the same issue we had before by shadowing the `styles` symbol with a local variable. What would be a good name for the local var?

src/styles/shadow.js
```diff
@@ -1,47 +1,77 @@
...
 				}
 			}
 		}
-	},
+	}
+
+	static [init] () {
+		super[init]?.();
+
+		if (!this[styles] || Object.hasOwn(this, initialized)) {
```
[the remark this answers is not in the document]
LeaVerou: This should be `this.styles`! No idea how it became a symbol. Might be Cursor being a bit too eager lol.

src/styles/global.js
```diff
@@ -50,17 +33,50 @@ export default {
...
-				if (!Self.roots.has(root)) {
-					adoptCSS(css, root);
-					Self.roots.add(root);
+		if (!this.globalStyles || Object.hasOwn(this, initialized)) {
+			return;
+		}
+
+		this[initialized] = true;
+
+		let supers = getSuperclasses(this, HTMLElement);
```
DmitrySharabin: I think we need `supers.push(this)` here as well?

LeaVerou: Yes. I'd swear I had done that. Strange.

src/mixins/form-associated.js
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		// Set current form value and update on change
+		this[internals].setFormValue(this[valueProp]);
+		let changeEventTarget = source || this;
+		changeEventTarget.addEventListener(changeEvent, () =>
+			this[internals].setFormValue(this[valueProp]));
+	}
+
+	static formAssociated = true;
```
DmitrySharabin: I noticed that this shadows a class's own `formAssociated` properties when this mixin is in use. That means that in `[onApply]()`, `config` is an empty object regardless. I might suggest the opposite. I don't know which part of the code should be fixed.

src/mixins/form-associated.js
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	attachInternals () {
+		return attachInternals(this);
+	}
+
+	[constructed] () {
+		let config = this.constructor[formAssociated] ?? this.constructor.formAssociated;
+		let { like, role, valueProp, changeEvent } = config;
+		let internals = this[internals] || this.attachInternals();
```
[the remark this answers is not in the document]
LeaVerou: This `internals` var is used differently: to store the internals object. We should replace its usage with `this[internals]`.

src/mixins/form-associated.js
LeaVerou: Actually, specifically for this one, I just had an idea that's easier.

PR #54, opened 2025-10-14
Fix issues with components initialization

DmitrySharabin:
- Apply mixins on initialization of the class itself and its children
- Run `init` hook on component mount
- `observedAttributes` getter should have context

### DIFF
--- a/src/Element.js
+++ b/src/Element.js
@@ -1,7 +1,7 @@
 /**
  * Base class for all elements
  */
-import mounted, { hasConnected } from "./mounted.js";
+import mounted from "./mounted.js";
 import props from "./props/defineProps.js";
 import formAssociated from "./form-associated.js";
 import events from "./events/defineEvents.js";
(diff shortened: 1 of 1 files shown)


src/mounted.js
```diff
@@ -15,7 +15,10 @@ export class MountedMixin extends HTMLElement {
 	}
 
 	/** Automatically gets called the first time the element is connected */
-	mounted () {}
+	mounted () {
+		// Stuff that runs once per element
+		this.constructor.hooks.run("init", this);
```
LeaVerou: We cannot depend that any class with this mixin has hooks

src/Element.js
```diff
@@ -38,24 +38,44 @@ const Self = class NudeElement extends HTMLElement {
...
-
 	static init () {
 		// Stuff that runs once per class
 		if (this[classInitialized]) {
 			return false;
 		}
 
+		// Every child class has to have the mounted mixin applied,
+		// but we don't want to share specific child class mixins with all other classes
+		this.mixins = [mounted];
```
LeaVerou: Why not keep this in the static declaration above then?

LeaVerou (changes requested):
The idea with the dynamic mixins array was that you'd build it in the static block, and then call `applyMixins()` like it was, not that you'd reintroduce doing it all via `init()`.

src/Element.js
LeaVerou: Nothing, you’re right!

src/mounted.js
DmitrySharabin: Should be fixed now. Found a better place (actually, where it was initially)—`connectedCallback()`.

PR #57, opened 2025-10-17
Refactor `ShadowStylesMixin` to use class mixins and eliminate the `setup` hook from `Element`

DmitrySharabin:
This mixin doesn't rely on the `init` hook anymore either.

### DIFF
--- a/src/styles/shadow.js
+++ b/src/styles/shadow.js
@@ -1,7 +1,9 @@
 /**
  * Mixin for adding shadow DOM styles
  */
-import { adoptCSS, fetchCSS, getSupers } from "./util.js";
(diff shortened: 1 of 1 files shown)


src/styles/shadow.js
LeaVerou: Presumably now that we’re using a class, we should be able to just call `super` and have it work instead of doing all this. Does it not?

DmitrySharabin: Nope, it doesn't work. `super` always refers to a class we extend our class (mixin) from (`HTMLElement`, in our case), not the class we apply the mixin to. So, this line will always be ignored (thanks to `?.()`). Without it, it always throws.

DmitrySharabin:
We don't need this anymore.

PR #58, opened 2025-10-17
Fix bugs and remove `debugger`

DmitrySharabin:


src/util/compose-functions.js
```diff
@@ -16,8 +16,7 @@ export function composeFunctions (fn1, fn2) {
 
 	if (isComposed) {
 		// A composed function was provided instead of the constituent
-		composedFn = isComposed;
-		if (!composedFn[functions]) debugger;
+		composedFn = fn1;
```
LeaVerou: While removing the debugger statement is obviously a good thing, I don't think this change is correct. `isComposed` in this case doesn't just act as a boolean, it contains the outcome of the composed function. We don't want to ever produce a second composed function for the same `fn1`, we want to reuse the existing one. Perhaps `composed` would have been a better name to make this clear.

PR #59, opened 2025-10-17
Refactor `GlobalStylesMixin` to use class mixins

DmitrySharabin:


src/styles/global.js
```diff
@@ -2,15 +2,20 @@
...
 import { adoptCSS, fetchCSS, getSupers } from "./util.js";
+import mounted from "../mounted.js";
 
-export default {
-	prepare () {
-		if (!this.globalStyles) {
+export const globalStylesFetched = Symbol("global styles fetched");
+
+export class GlobalStylesMixin extends HTMLElement {
+	static mixins = [mounted];
```
LeaVerou: One thing to check is, how do we handle the case where two mixins include the same mixin? We should not be including it twice (or, if we are, the result should not be observably different).

src/styles/global.js
```diff
@@ -2,15 +2,20 @@
...
+
+	/** Automatically runs once per class the first time an instance is connected */
+	static mounted () {
+		if (!this.globalStyles || Object.hasOwn(this, globalStylesFetched)) {
 			return;
 		}
 
 		let supers = getSupers(this);
```
LeaVerou: Same comment as the other PR: class inheritance should be able to take care of this now. Does it not?

PR #61, opened 2025-10-29
Some fixes to the `class-mixins` branch

DmitrySharabin:
Fix imports and exports, function calls, and some name collisions.

### DIFF
--- a/src/mixins/apply.js
+++ b/src/mixins/apply.js
@@ -1,7 +1,7 @@
 import { copyProperties } from "../util/copy-properties.js";
 
 export function applyMixins (Class = this, mixins = Class.mixins) {
-	if (Object.hasOwn(Class, "mixinsActive") || this && !Object.hasOwn(this, "mixins")) {
+	if (Object.hasOwn(Class, "mixinsActive") || !mixins?.length) {
 		return;
 	}
 


src/mixins/apply.js
```diff
@@ -1,7 +1,7 @@
 import { copyProperties } from "../util/copy-properties.js";
 
 export function applyMixins (Class = this, mixins = Class.mixins) {
-	if (Object.hasOwn(Class, "mixinsActive") || !Object.hasOwn(this, "mixins")) {
+	if (Object.hasOwn(Class, "mixinsActive") || this && !Object.hasOwn(this, "mixins")) {
```
DmitrySharabin: Otherwise, `Object.hasOwn(undefined)` throws.

src/styles/util.js
```diff
@@ -1,4 +1,4 @@
-export * from "./util/adopt-css.js";
-export * from "./util/fetch-css.js";
-export * from "./util/get-symbols.js";
-export * from "./util/get-supers.js";
+export * from "../util/adopt-css.js";
+export * from "../util/fetch-css.js";
+export * as getSymbols from "../util/get-symbols.js";
```
DmitrySharabin: Probably, there should be a better way to export the proxy object as a named export so that it can be imported like all other utils, i.e., `import { getSymbols } from "..."`. The change in `nude-element.js` is related to this change and to the fact that there's no named export in `get-symbols.js`.

src/util/extend.js
```diff
@@ -9,9 +9,9 @@ export const sideEffects = Symbol("Side effects");
 export const mutable = Symbol("Mutable");
 
 export function extend (body, ...sideEffects) {
-	let mutable = body[sideEffects] ? body : body[mutable];
+	let Mutable = body[sideEffects] ? body : body[mutable];
```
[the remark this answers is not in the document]
DmitrySharabin: Let's go with `mutableFn`

src/styles/global.js
```diff
@@ -1,10 +1,7 @@
 /**
  * Mixin for adding light DOM styles
  */
-import { getSupers } from "../util/get-supers.js";
-import { adoptCSSRecursive } from "../util/adopt-css.js";
-import { fetchCSS } from "../util/fetch-css.js";
-import getSymbols from "../util/get-symbols.js";
+import { adoptCSSRecursive, fetchCSS, getSupers, getSymbols } from "./util.js";
```
DmitrySharabin: It's just a proposal to simplify imports the same way we already do in `shadow.js`. No issues here.

src/mixins/apply.js
DmitrySharabin: Probably, there should be something like this: `if (Object.hasOwn(Class, "mixinsActive") || !mixins?.length) { ... }`?

DmitrySharabin: Committed that check. Seems reasonable to me and doesn't break anything (at least in my tests 😊).

LeaVerou (approved):

PR #65, opened 2025-11-23
Some fixes to the updated code

DmitrySharabin:


src/mixins/form-associated.js
```diff
@@ -70,14 +70,14 @@ export const Mixin = (Super = HTMLElement) => class FormAssociated extends Super
...
+	static [onApply] (Class) {
+		let config = Class[formAssociated] || Class.formAssociated;
 		config = !config || typeof config !== "object" ? {} : config;
 
-		this[formAssociated] = getOptions(defaultOptions, config);
+		Class[formAssociated] = getOptions(defaultOptions, config);
 
-		delegate({
-			properties: this[formAssociated].properties,
+		delegate.call(Class, {
```
DmitrySharabin: Inside `delegate()`, we use `this` in the getters we define, so we need `Class` as a context.

src/util/extend-object.js
```diff
@@ -68,7 +68,10 @@ export function extendObject (target, source, options = {}) {
 	let skippedProperties = new Set(options.skippedProperties || []);
 	let sourceDescriptors = Object.getOwnPropertyDescriptors(source);
 
-	for (let prop in sourceDescriptors) {
+	for (let prop of [
+		...Object.keys(sourceDescriptors),
+		...Object.getOwnPropertySymbols(sourceDescriptors),
+	]) {
```
DmitrySharabin: Otherwise, we miss symbols.

src/mixins/form-associated.js
```diff
@@ -70,14 +70,14 @@ export const Mixin = (Super = HTMLElement) => class FormAssociated extends Super
 
 	static formAssociated = true;
 
-	static [onApply] () {
-		let config = this[formAssociated] || this.formAssociated;
+	static [onApply] (Class) {
```
LeaVerou: Why? This is a static method, so this should resolve to the class anyway

src/util/extend-object.js
LeaVerou: getOwnPropertyDescriptors should return symbols too (unlike getOwnPropertyNames)

DmitrySharabin: It does, but `for…in` doesn't iterate over them

src/mixins/form-associated.js
DmitrySharabin: I've found the "real" issue—we called it incorrectly: we should've passed `Class` as a context, not a parameter. Fixed

src/mixins/form-associated.js
```diff
@@ -48,7 +48,7 @@ export const Mixin = (Super = HTMLElement) => class FormAssociated extends Super
 	[constructed] () {
 		let config = this.constructor[formAssociated] ?? this.constructor.formAssociated;
 		let { like, role, valueProp, changeEvent } = config;
-		let internals = this[internals] || this.attachInternals();
+		this[internals] ??= this.attachInternals();
```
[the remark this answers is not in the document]
DmitrySharabin: This covers cases where `this.attachInternals` was already called before `[constructed]()` is executed, e.g., in a class constructor (like we do in `ColorElement`).

src/util/delegate.js
```diff
@@ -17,7 +17,7 @@ export function delegate ({from, to, properties, descriptors}) {
 			configurable: true,
 		};
 
-		if (sourceDescriptor.writable || sourceDescriptor.set) {
+		if (sourceDescriptor?.writable || sourceDescriptor?.set) {
```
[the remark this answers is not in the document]
DmitrySharabin: Yes, it's one of the default properties:

Image

src/util/extend-object.js
LeaVerou: Ah, true, `for .. in` skips symbols. `Reflect.ownKeys()` should give you all of them.

src/mixins/styles/global.js
```diff
@@ -61,7 +61,7 @@ export const Mixin = (Super = HTMLElement) => class GlobalStyles extends Super {
 
 		let Super = Object.getPrototypeOf(this);
 
-		if (Super[self]) {
+		if (!Super[self]) {
```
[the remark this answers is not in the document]
DmitrySharabin: Okay, now I got it. So, to make the `GlobalStyles` mixin work, a class should import the `self` symbol from `global.js` and add a static property `[self]` with any (?) value (I added `static [self] = self`). Is that correct?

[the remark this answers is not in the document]
LeaVerou: No, applying the mixin should add it automatically.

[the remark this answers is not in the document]
DmitrySharabin: Well, this isn't happening. I'll take a closer look

src/mixins/form-associated.js
DmitrySharabin: Here (you've already moved my change to your PR):
image

PR #66, opened 2025-12-08
Another attempt to refactor

LeaVerou:


DmitrySharabin (changes requested):
See comments

src/events/defineEvents.js
```diff
@@ -1,152 +1,25 @@
 import defineProps from "../props/defineProps.js";
```
DmitrySharabin:
```suggestion
import * as defineProps from "../props/defineProps.js";
```

src/Element.js
```diff
@@ -65,22 +73,65 @@ const Self = class NudeElement extends HTMLElement {
...
+			extend(this, plugin.members);
 		}
 
-		this.hooks.run("setup", this);
+		if (plugin.membersStatic) {
+			extend(this, plugin.membersStatic);
+		}
 
-		return (this[classInitialized] = true);
+		plugin.hooks.add(this.hooks);
```
DmitrySharabin: Here, if I'm not mistaken, we expect `plugin.hooks` to be an instance of the `Hooks` class, but it's not, and that assumption causes all code that depends on it to fail.
I suppose this should happen automatically, but I couldn't find an appropriate place for it. Yet.

src/events/onprops.js
```diff
@@ -0,0 +1,70 @@
+
+/**
+ * Add on* props for UI events, just like native UI events
+ */
+
+import getSymbols from "../util/symbols.js";
```
DmitrySharabin: I know that we can rename default exports the way we like. Just want to make sure it's intentional. In `symbols.js`, we call it `newSymbols`.

DmitrySharabin: We seem to name it `getSymbols` all over the code base, so it might be intentional? 🤔

LeaVerou: Not intentional haha. I started by calling it `getSymbols`, then decided `newSymbols` was a better name, and clearly now I just can't remember 🤣

src/events/propchange.js
```diff
@@ -0,0 +1,47 @@
...
+ * Implement propchange events:
+ * Events that fire when a specific prop changes
+ */
+
+import getSymbols from "../util/symbols.js";
+const { propchange } = getSymbols;
+
+export const hooks = {
+	first_constructor_static () {
+		let propchangeEvents = Object.entries(this[events])
```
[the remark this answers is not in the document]
LeaVerou: It should definitely be a known symbol, yes. Added by `events.js`. `eventProps` is not used in this file at all.

src/util/unused/copy-properties.js
```diff
@@ -1,4 +1,4 @@
-import { composeFunctions } from "./compose-functions.js";
+import { composeFunctions } from "./unused/compose-functions.js";
```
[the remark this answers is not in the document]
LeaVerou: But it _is_ in `unused`?

src/Element.js
```diff
@@ -16,11 +19,8 @@ const Self = class NudeElement extends HTMLElement {
 	constructor () {
 		super();
 
-		if (!this.constructor[classInitialized]) {
-			this.constructor.init();
-		}
-
-		this.constructor.hooks.run("start", this);
+		this.constructor.hooks.run("constructor-static", this.constructor);
+		this.constructor.hooks.run("constructor", this);
```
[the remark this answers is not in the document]
LeaVerou: Damn. We should fix it in Hooks, `getCanonicalName` could check if the hook is a reserved word and prepend some kind of prefix. Or use a `Map` instead of an object literal where all values are fine.

src/form-behavior/index.js
```diff
@@ -0,0 +1,9 @@
+import * as base from "./base.js";
+import * as like from "./like.js";
+import * as delegate from "./delegate.js";
+
+export function setup () {
+	this.addPlugin(base);
+	this.addPlugin(like);
+	this.addPlugin(delegate);
+}
```
DmitrySharabin: Shouldn't this be written the following way?

```js
export default {
	setup () {
		this.addPlugin(base);
		this.addPlugin(like);
		this.addPlugin(delegate);
	},
}
```

src/props/base.js
```diff
@@ -0,0 +1,87 @@
...
+	first_constructor_static,
+
+	first_connected () {
+		this.constructor.props.initializeFor(this);
+	},
+};
+
+export const members = {
+	// Internal prop values
+	props () {
```
DmitrySharabin: I wonder if these methods should be getters? We seem to work with them as if they are. See L60, for example, where we access the `props` property to get the default value of `def`.

DmitrySharabin: No, they shouldn't (I'm gonna answer myself 😆). I noticed, though, that this member shadows the class's static props somehow. 🤷‍♂️

src/props/base.js
```diff
@@ -0,0 +1,87 @@
...
+	// FIXME how to combine with existing observedAttributes?
+	if (!Object.hasOwn(this, "observedAttributes")) {
+		Object.defineProperty(this, "observedAttributes", {
+			get: () => this.props.observedAttributes,
+			configurable: true,
+		});
+	}
+
+	if (this.props) {
+		this.defineProps(this);
```
DmitrySharabin:
```suggestion
		this.defineProps(this.props);
```

src/props/base.js
LeaVerou: Why? I was going to say that it should!

src/plugins.js
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if (plugin.dependencies) {
+		for (let dependency of plugin.dependencies) {
+			addPlugin(Class, dependency);
+		}
+	}
+
+	if (plugin.members) {
+		extend(Class, plugin.members);
```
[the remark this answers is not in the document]
LeaVerou: Should be fixed now!

src/mixins/hooks.js
```diff
@@ -49,12 +66,59 @@ export default class Hooks {
...
+
+	/**
+	 * Like run(), but only executes the callback once per context
+	 * @param {*} env
+	 */
+	runOnce (env) {
+		for (let callback of this) {
+			let context = env?.context ?? env;
+
+			if (this.contexts.has(context)) {
```
DmitrySharabin: Because of this, we'll run only the first callback (of the first added plugin). I believe it's props. We should probably track not only context but also the plugin the hook belongs to.

src/form-behavior/base.js
```diff
@@ -0,0 +1,37 @@
...
+
+		const env = {context: this, formBehavior: def};
+		this.hooks.run("define-form-behavior", env);
+
+		this[formBehavior] ??= {};
+		Object.assign(this[formBehavior], env.formBehavior);
+	},
+};
+
+export default {dependencies, setup, membersStatic};
```
DmitrySharabin:
```suggestion
export default { dependencies, setup, hooks, membersStatic };
```

src/util/delegate.js
```diff
@@ -0,0 +1,28 @@
...
+		let descriptor = {
+			get () {
+				return this[to][prop];
+			},
+			...sourceDescriptor,
+			configurable: true,
+		};
+
+		if (sourceDescriptor.writable || sourceDescriptor.set) {
+			descriptor.set = function (value) {
```
DmitrySharabin:
```suggestion
			delete descriptor.value;
			delete descriptor.writable;

			descriptor.set = function (value) {
```

Otherwise, some descriptors might throw on line 26. For example, `checkValidity` is writable and has a value. After adding accessors, we make the descriptor invalid. 

Image

src/internals/base.js
```diff
@@ -0,0 +1,39 @@
...
+
+import symbols from "../util/symbols.js";
+import { defineLazyProperty } from "../util/lazy.js";
+
+const { internals } = symbols.known;
+const { attachInternals } = HTMLElement.prototype;
+
+export const members = {
+	attachInternals () {
+		if (this[internals] !== undefined) {
```
DmitrySharabin: This line throws a stack overflow error. 🤷‍♂️

src/events/onprops.js
```diff
@@ -0,0 +1,74 @@
...
+
+			this.hooks.add("first-connected", function firstConnected () {
+
+			});
+		}
+	},
+
+	first_connected () {
+		// Deal with existing values
+		if (!this[eventProps]) {
```
DmitrySharabin:
```suggestion
		if (!this.constructor[eventProps]) {
```

src/events/onprops.js
```diff
@@ -0,0 +1,74 @@
...
+			return;
+		}
+
+		for (let name in this[eventProps]) {
```
DmitrySharabin:
```suggestion
		for (let name in this.constructor[eventProps]) {
```

src/events/propchange.js
```diff
@@ -0,0 +1,51 @@
...
+				else {
+					throw new TypeError(`No prop named ${propName} in ${this.name}`);
+				}
+			}
+		}
+	},
+
+	first_connected () {
+		// Often propchange events have already fired by the time the event handlers are added
+		for (let eventName in this[propchange]) {
```
DmitrySharabin:
```suggestion
		for (let eventName in this.constructor[propchange]) {
```

src/mixins/hooks.js
LeaVerou: Ah yes, it should be running each callback once. 

This is a nice self-contained intermediate thing you could work on if you want!

src/events/propchange.js
```diff
@@ -0,0 +1,51 @@
...
+			let propName = this[propchange][eventName];
+			let value = this[propName];
+
+			if (value !== undefined) {
+				this.props.firePropChangeEvent(this, eventName, {
```
DmitrySharabin:
```suggestion
				this.constructor.props.firePropChangeEvent(this, eventName, {
```

src/events/propchange.js
```diff
@@ -0,0 +1,51 @@
...
+					name: propName,
+					prop: this.props.get(propName),
```
DmitrySharabin:
```suggestion
					prop: this.constructor.props.get(propName),
```

src/events/propchange.js
LeaVerou: No I think this is supposed to be the instance object. This is not new code.

src/internals/base.js
LeaVerou: - The `internals` plugin is separate from form functionality (that's `form-behavior`). Not all `ElementInternals` functionality is about forms.
- I renamed the variable
- We do want to overwrite `this.attachInternals()` so that whatever calls it we still get a reference to the internals.
- I'll think some more about the getter issue

src/events/propchange.js
```diff
@@ -0,0 +1,51 @@
...
+					(prop.eventNames ??= []).push(eventName);
+				}
+				else {
+					throw new TypeError(`No prop named ${propName} in ${this.name}`);
+				}
+			}
+		}
+	},
+
+	first_connected () {
```
DmitrySharabin:
```suggestion
	connected () {
```

We should fire such events for every connected instance since they might have event handlers.

DmitrySharabin: Our color elements fire `onPROPchange` events during initialization. They won't do that if we run the code in this hook once the first element is connected.
For example, we won't see the space picker value in the `<output>` element if we don't run this hook for every element:
```html
<space-picker onspacechange="this.nextElementSibling.textContent = this.value"></space-picker>
<output></output>
```

src/internals/base.js
```diff
@@ -0,0 +1,40 @@
...
+import symbols from "../util/symbols.js";
+import { defineLazyProperty } from "../util/lazy.js";
+
+const { internals } = symbols.known;
+const _attachInternals = HTMLElement.prototype.attachInternals;
+
+export const members = {
+	attachInternals () {
+		let descriptor = Object.getOwnPropertyDescriptor(this, internals);
+		if (descriptor.value) {
```
DmitrySharabin:
```suggestion
		if (descriptor?.value) {
```

src/internals/base.js
DmitrySharabin: Proposed one tiny change (`descriptor` might be `undefined`, so we need to consider it while accessing its `value` property).

src/events/propchange.js
DmitrySharabin: But if we don't rename it, the events fired during initialization are not handled. Hm. I'm a bit confused now.

DmitrySharabin: > `first_connected` should run for every instance, but only the first time it's connected, whereas `connectedCallback()` (and thus, the `connected` hook) runs every time.
> 
> E.g. consider the following:
> 
> ```js
> let el = document.createElement("foo-bar");
> document.body.append(el); // connected, first_connected
> el.remove();
> document.body.append(el); // connected
> ```

That makes sense. Now I got it. Thanks!

LeaVerou: > > Which events fired during initialization?
> 
> `valuechange` and `spacechange`

Another issue is described by the comment below: callacks added to the hook after it ran will not be executed. We *can* make it so that they are immediately executed in that case, but I don't know if that's desirable for every case.

src/form-behavior/base.js
```diff
@@ -0,0 +1,45 @@
...
+		if (!this.formBehavior) {
+			this.formBehavior = true;
+		}
+
+		const env = {context: this, formBehavior: def};
+		this.hooks.run("define-form-behavior", env);
+
+		this[formBehavior] ??= {};
+		Object.assign(this[formBehavior], env.formBehavior);
+	},
```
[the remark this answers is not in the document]
LeaVerou: I would swear I had dealt with this before. Turns out I had accidentally mass-replaced `formAssociated` with `formBehavior. Just pushed a fix.

src/form-behavior/base.js
```diff
@@ -0,0 +1,45 @@
...
+
+export const membersStatic = {
+	defineFormBehavior (def = this[formBehavior] ?? this.formBehavior) {
+		if (!def) {
+			return;
+		}
+
+		if (!this.formAssociated) {
+			this.formAssociated = true;
+		}
```
DmitrySharabin: Does this set the static `formAssociated` property? Because it doesn't seem to fix the bug:

Image

LeaVerou: It should, yes. But you can check.

src/form-behavior/base.js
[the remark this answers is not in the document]
LeaVerou: I would swear I had dealt with this before. Turns out I had accidentally mass-replaced `formAssociated` with `formBehavior. Just pushed a fix.

src/form-behavior/like.js
```diff
@@ -0,0 +1,31 @@
+import symbols from "../util/symbols.js";
+import { resolveValue } from "../util.js";
+import { getRole } from "./role.js";
+
+const { formBehavior, internals } = symbols.known;
+
+export const hooks = {
+	first_connected () {
+		let { like, role, valueProp = "value", changeEvent = "input" } = this.constructor[formBehavior];
```
DmitrySharabin:
```suggestion
		if (!this.constructor[formBehavior]) {
			return;
		}

		let { like, role, valueProp = "value", changeEvent = "input" } = this.constructor[formBehavior];
```

We should do the stuff only with suitable instances.

src/shadow/base.js
```diff
@@ -0,0 +1,44 @@
...
+
+import symbols from "../util/symbols.js";
+import { defineLazyProperty } from "../util/lazy.js";
+
+const { shadowRoot, shadowRootOptions } = symbols.known;
+const { attachShadow } = HTMLElement.prototype;
+
+export const members = {
+	attachShadow (options = this.constructor[shadowRootOptions] ?? this.constructor.shadowRoot) {
+		if (this[shadowRoot] !== undefined) { // We want to include null
```
DmitrySharabin: This might also suffer from an infinite recursion.

LeaVerou: No because we don't call `attachShadow()` automatically when `[shadowRoot]` is accessed

src/plugins.js
```diff
@@ -0,0 +1,56 @@
...
+		extend(Class.prototype, plugin.provides);
+	}
+
+	if (plugin.providesStatic) {
+		extend(Class, plugin.providesStatic);
+	}
+
+	Class.hooks.add(plugin.hooks);
+
+	plugin.setup?.call(Class);
```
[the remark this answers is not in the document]
DmitrySharabin: Unfortunately, it's not. But maybe it should be earlier, not at the end of the method? 🤔

Let's consider this a heads-up so you can take a look when you have some cycles.

src/elements/base.js
```diff
@@ -0,0 +1,75 @@
...
+	if (multiple) {
+		return Array.from(root.querySelectorAll(selector));
+	}
+	else {
+		return root.querySelector(selector);
+	}
+}
+
+export const hooks = {
+	connected () {
```
DmitrySharabin:
```suggestion
	first_constructor_static () {
		if (this.elements) {
			this.defineElements();
		}
	},

	connected () {
```

src/Element.js
```diff
@@ -45,42 +38,66 @@ const Self = class NudeElement extends HTMLElement {
...
-			defineFormAssociated(this);
-		}
+	static hasPlugin (plugin) {
+		return hasPlugin(this, plugin);
+	}
 
-		if (this.styles) {
-			defineMixin(this, shadowStyles);
-		}
+	static addPlugin (plugin) {
```
DmitrySharabin: I think it would be helpful to support adding a list (an array) of plugins. I already needed it.

src/elements/base.js
```diff
@@ -0,0 +1,75 @@
...
+			this[elements] = {};
+			defineLazyProperty(this.prototype, elements, {
+				value: {}, // mainly to ensure the getter doesn't get overwritten on the prototype
+				get () {
+					let ret = {};
+					for (let name in this.constructor[elements]) {
+						defineLazyProperty(ret, name, {
+							get () {
+								return getElement(this, this.constructor[elements][name]);
+							},
```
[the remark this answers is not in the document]
DmitrySharabin: Make the props accessible in `for .. in` loops.

src/util.js
```diff
@@ -2,8 +2,5 @@ export * from "./util/resolve-value.js";
 export * from "./util/is-class.js";
 export * from "./util/is-subclass-of.js";
 export * from "./util/lazy.js";
-export * from "./util/extend-class.js";
-export * from "./util/copy-properties.js";
-export * from "./util/compose-functions.js";
-export * from "./util/reversible-map.js";
 export * from "./util/pick.js";
+export * from "./util/get-supers.js";
```
DmitrySharabin:
```suggestion
export * from "./util/super.js";
```

src/slots/base.js
```diff
@@ -0,0 +1,30 @@
...
+		if (!(this[slots] instanceof Slots)) {
+			this[slots] = new Slots(this, def);
+		}
+		else if (this[slots] === def) {
+			// Nothing to do here
+			return;
+		}
+
+		// New slots to add
+		this[slots].add(def);
```
DmitrySharabin:
```suggestion
		this[slots].define(def);
```

I think this method should be here (at least there's no `add()` method, and `define ()` does add slots).

src/mo2/unobserve.js
```diff
@@ -0,0 +1,33 @@
...
+		for (let [ref, options] of this.observations.entries()) {
+			let mutationTarget = ref.deref();
+
+			if (mutationTarget && options) {
+				this.observe(mutationTarget, options);
+			}
+			else {
+				// Possibly garbage collected
+				this.observations.delete(ref);
+			}
```
DmitrySharabin: Based on your previous comment, this should be redundant as well, no? In any case, it won't do anything since ref is an unwrapped value here.

src/slots/named-manual.js
```diff
@@ -5,104 +5,53 @@
...
-					assign(node, slots);
-				}
-
-				for (let node of nodesToUnassign) {
-					unassign(node, slots);
-				}
-			});
+export const hooks = {
+	connected () {
+		if (this[slots].shadowRoot?.slotAssignment !== "manual") {
```
DmitrySharabin:
```suggestion
		if (this[slots]?.shadowRoot?.slotAssignment !== "manual") {
```

In my experiments, for elements without slots, this line might throw an error, so I'd make it a bit more error-proof.

DmitrySharabin:
### A list of issues to consider before merging (🚫 — blocking, ℹ — non-blocking):
ℹ https://github.com/nudeui/element/pull/66#discussion_r2624024189

ℹ https://github.com/nudeui/element/pull/66#discussion_r2624719695

🚫 https://github.com/nudeui/element/pull/66#discussion_r2624800925

🚫 Infinite recursion with the `*` (universal) hook: https://github.com/nudeui/element/blob/4cf26a9ad2f25e79494aa09789f6008644189126/src/hooks.js#L95-L98

src/extensible.js
```diff
@@ -0,0 +1,86 @@
+import Hooks from "./hooks.js";
+import { hasPlugin, addPlugin } from "./plugins.js";
+import { getSuper } from "./util/super.js";
+import symbols from "./util/symbols.js";
+
+const {hooks, plugins, initialized} = symbols.new;
+
+export default function makeExtensible (Class) {
+	Object.defineProperty(Class, "hooks", {
```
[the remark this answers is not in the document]
LeaVerou: It's still essentially lazy, the accessor creates it lazily and returns it. The difference is that previously we were overwriting it with a regular data property, so if we had a hierarchy like C > B > A, once we rewrote `B.hooks` to a data property, it would shadow the accessor on `A.hooks` so `C.hooks` would just be pointing to `B.hooks` like a regular data property. Here we keep the accessor at all times, so that doesn't happen. Or at least that's the idea!

src/internals/base.js
```diff
@@ -0,0 +1,50 @@
...
+function getOwnValue (object, name) {
+	let descriptor = Object.getOwnPropertyDescriptor(object, name);
+	return descriptor?.value;
+}
+
+export const provides = {
+	attachInternals () {
+		if (getOwnValue(this, internals)) {
+			return this[internals];
+		}
```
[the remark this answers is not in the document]
LeaVerou: Gah, yes.

src/styles/shadow.js
```diff
@@ -1,47 +1,45 @@
...
-	setup () {
+const { shadowStyles } = symbols.new;
+const { shadowRoot } = symbols.known;
+
+export const hooks = {
+	// Initiate fetching when the first element is constructed
+	first_constructor_static () {
 		if (!this.styles) {
 			return;
 		}
```
DmitrySharabin: I wonder if we still need this check

src/styles/shadow.js
```diff
@@ -1,47 +1,45 @@
...
+			// Get fetched styles from this and all superclasses that define any
+			// We're doing a bit of duplicate work here, but that's ok because fetchCSS() caches results anyway
+			this[shadowStyles] = getAllValues(this, "styles").flat().map(localUrl => fetchCSS(localUrl, this.url));
 		}
 	},
-	async init () {
-		if (!this.shadowRoot) {
+
+	async first_connected () {
+		if (!this.shadowRoot && !this[shadowRoot]) {
```
DmitrySharabin: I don't understand why we need to check for presence of `this[shadowRoot]`. We seem don't depend on it anywhere in the hook.

src/styles/global.js
```diff
@@ -1,45 +1,37 @@
...
-export default {
-	prepare () {
+const { globalStyles, roots } = symbols.new;
+
+export const hooks = {
+	// Initiate fetching when the first element is constructed
+	first_constructor_static () {
 		if (!this.globalStyles) {
 			return;
 		}
```
DmitrySharabin: Do we still need this check? Even if `globalStyles` is somewhere in the prototype chain, we won't use it if it's not an own property of `this`.

src/styles/global.js
```diff
@@ -1,45 +1,37 @@
...
-			if (
-				Object.hasOwn(Class, "globalStyles") &&
-				!Object.hasOwn(Class, "fetchedGlobalStyles")
-			) {
-				// Initiate fetching when the first element is constructed
-				let styles = (Class.fetchedGlobalStyles = Array.isArray(Class.globalStyles)
-					? Class.globalStyles.slice()
-					: [Class.globalStyles]);
-				Class.roots = new WeakSet();
+		this[roots] = new WeakSet();
```
DmitrySharabin: As I can see, this method is only used in the `connected` hook. Is there a reason not to define it there (like `this[roots] ??= new WeakSet();`) a bit closer to the code where it's actually used? 🤔

src/styles/shadow.js
LeaVerou: Ah good point, here we don't

src/styles/shadow.js
LeaVerou: If the shadow plugin is used, then `this[shadowRoot]` would contain closed roots, but `this.shadowRoot` would be empty. But we don't want to *depend* on the shadow plugin.

src/plugins/util/super.js
```diff
@@ -0,0 +1,97 @@
...
+export function getSuperMember (obj, name) {
+	let thisMember = Object.getOwnPropertyDescriptor(obj, name);
+	let Super = obj;
+
+	while (Super = getSuper(Super, name)) {
+		if (!Super) {
+			return undefined;
+		}
+
+		let superMember = Object.getOwnPropertyDescriptor(Super, name);
```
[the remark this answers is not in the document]
LeaVerou: I _think_ when we pass an instance as `obj`, `getSuper()` returns the prototype already so that shouldn't be needed.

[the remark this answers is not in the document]
LeaVerou: Just pushed a potential fix!

src/plugins/index.js
```diff
@@ -0,0 +1,7 @@
+export { default as makeExtensible } from "./extensible.js";
+export * from "./extensible.js";
+export * from "../util/own.js";
```
DmitrySharabin:
```suggestion
export * from "./util/own.js";
```

src/plugins/util/own.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {*} obj
+ * @param {*} name
+ * @param {*} options
+ */
+export function defineOwnProperty (obj, name, options = {}) {
+	if (typeof options === "function") {
+		options = { get: options };
+	}
+
+	let _name = options.internal ?? Symbol(name);
```
DmitrySharabin: If `name` is already a symbol, this line will throw. We should probably add an explicit check. Or `name` shouldn't be a symbol in the first place? 🤔

src/styles/global.js
```diff
@@ -1,18 +1,17 @@
...
-const { globalStyles, roots } = symbols.new;
+const { globalStyles } = symbols.new;
 
 export const hooks = {
 	// Initiate fetching when the first element is constructed
 	first_constructor_static () {
 		if (Object.hasOwn(this, "globalStyles")) {
 			// Get fetched styles from this and all superclasses that define any
-			// We're doing a bit of duplicate work here, but that's ok because fetchCSS() caches results anyway
-			this[globalStyles] = getAllValues(this, "globalStyles").flat().map(localUrl => fetchCSS(localUrl, this.url));
```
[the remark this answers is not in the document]
LeaVerou: Gah, nice catch. Will think about it.

src/element/index.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Base class for all elements
+ */
+
+export { default as ElementFactory } from "./factory.js";
+export { default as elementMembers } from "./members.js";
+export { default as Element } from "./element.js";
```
DmitrySharabin:
```suggestion
export { default as Element } from "./Element.js";
```

src/element/index.js
```diff
@@ -0,0 +1,8 @@
...
+export { default as default } from "./element.js";
```
DmitrySharabin:
```suggestion
export { default as default } from "./Element.js";
```

src/plugins/styles/global.js
```diff
@@ -0,0 +1,44 @@
...
+	connected_apply_style ({ roots, options }) {
+		if (options.roots.has("global")) {
+			for (let root of getRoots(this)) {
+				roots.add(root);
+			}
+		}
+	},
+
+	define_style ({ style }) {
+		if (style.global) {
```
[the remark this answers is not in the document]
LeaVerou: Good catch! Yes, that's the problem. Let me think how to best solve it…

DmitrySharabin (approved):
L(exceptionally)GTM!

PR #74, opened 2025-12-15
[shadow] Some fixes

DmitrySharabin:
- Don’t shadow `attachShadow` from `HTML.prototype`
- Avoid infinite recursion
- Correctly run hooks

### DIFF
--- a/src/shadow/base.js
+++ b/src/shadow/base.js
@@ -24,7 +24,7 @@ export const provides = {
 
 		try {
 			this[shadowRoot] = _attachShadow.call(this, options);
-			this.hooks.run("shadow-attached", {context: this, shadowRoot});
+			this.constructor.hooks.run("shadow-attached", {context: this, shadowRoot});
 		}
 		catch (error) {
 			this[shadowRoot] = null;


src/shadow/base.js
```diff
@@ -6,23 +6,24 @@ import symbols from "../util/symbols.js";
...
+const _attachShadow = HTMLElement.prototype.attachShadow;
 
 export const provides = {
 	attachShadow (options = this.constructor[shadowRootOptions] ?? this.constructor.shadowRoot) {
-		if (this[shadowRoot] !== undefined) { // We want to include null
-			return this[shadowRoot];
+		let descriptor = Object.getOwnPropertyDescriptor(this, shadowRoot);
+		if (descriptor?.value) {
+			return descriptor.value;
 		}
```
DmitrySharabin: The same fix we applied in the `internals` plugin.

DmitrySharabin: Otherwise, we suffer from infinite recursion if someone calls `this.attachShadow()` inside `constructor`.

DmitrySharabin: By the way, this is how the refactored `ColorElement` class looks now. I think it's beautiful. 🤩

image

PR #75, opened 2025-12-16
Make the elements plugin work

DmitrySharabin:


src/elements/base.js
```diff
@@ -24,7 +24,13 @@ function getElement (host, options) {
...
 
 export const hooks = {
-	connected () {
+	first_constructor_static () {
+		if (this.elements) {
+			this.defineElements();
+		}
+	},
+
+	constructed () {
```
DmitrySharabin: Interestingly, the code of this hook should run at a different time depending on cases:
- It should run before we begin to work with internals, form-associated, events, etc
- But at the same time, it should run after an element is connected, so all other custom elements used inside the element are connected and ready to be queried

I can't think of a hook that we should use here. I'm puzzled. 😕

DmitrySharabin: `first_connected` might be a better choice? Hm...

Issue #79, opened 2025-12-28
[Props] Use signals


LeaVerou:
Currently, we do some pretty awkward things to figure out dependencies (poorly). We should instead use signals, and expose them for authors to use as well.
Possibly [tiny-signals](https://github.com/jsebrech/tiny-signals/) or something like it.

PR #91, opened 2026-03-19
Use signals for props reactivity (#79)

LeaVerou:
## Summary

Replaces fragile regex-based static dependency inference (`inferDependencies()`) and manual dependency cascade (`Props.updateDependents()` / topological sort) with a signals-based auto-tracking system.

- **`src/signals.js`** (new): minimal `Signal` and `Computed` classes. `Computed` auto-discovers dependencies by intercepting `.value` reads during its compute function — no explicit dep declaration needed. Re-tracks each recomputation, so dynamic/conditional deps work.
- **`Prop.js`**: adds `getSignal(element)` returning a `Signal` (plain props) or `Computed` (computed props) per element. Reads go through `signal.value` for auto-tracking; writes update `signal.value` to propagate automatically. Removes `dependencies`, `dependsOn()`, `update()`.
- **`Props.js`**: removes `dependents` graph, `updateDependents()`, and topological sort. `propChanged()` now only fires events — signals handle propagation.
- **`util.js`**: removes `inferDependencies()` and `sortObject()`.
- **`test/Prop.js`**: removes dep inference tests, updates defaults test.

Authors writing computed props just use `this.foo` in `get` — deps are discovered at runtime with no extra configuration.

## Test plan

- [x] All existing tests pass (`npm test`)
- [x] Computed props update when dependencies change
- [x] Dynamic/conditional deps (`if (this.foo) return this.bar`) work correctly
- [x] `propchange` events still fire for both plain and computed props
- [x] Attribute reflection still works for plain props
- [x] Function defaults that reference other props update reactively


LeaVerou:
## Code Review: Issues Found

I've reviewed the signals-based reactivity rewrite. The overall direction is solid — replacing fragile regex-based inference with auto-tracking is the right call. However, there are several issues that need to be addressed before merging:

### 1. Critical: `convert` reactivity is broken

The old `Prop.update()` re-ran `convert` when dependencies changed:

```js
if (this.spec.convert && oldValue !== undefined) {
    let value = this.spec.convert.call(element, oldValue);
    this.set(element, value, { source: "convert", oldValue });
}
```

This is completely removed. `convert` still runs inside `set()` when a value is directly set, but if `convert` references other props (e.g. `convert(v) { return v * this.scale }`), changing `scale` won't re-trigger conversion. The old system tracked `convert` dependencies via `inferDependencies(spec.convert)`.

### 2. Critical: Synchronous recursive recomputation causes glitching

In `Computed.#compute()`, when a dependency changes, the subscriber calls `this.#compute()` immediately:

```js
dep.subscribe(() => {
    this.#dirty = true;
    this.#compute();
});
```

For a diamond dependency graph (A→B, A→C, B→D, C→D), changing A will compute D **twice** — once when B updates and once when C updates. The first computation sees stale C. The old topological sort ensured each prop updated exactly once in the correct order.

### 3. `element.props[this.name]` bypass in computed subscriber

In `getSignal()`, the computed subscriber writes directly to `element.props`:

```js
signal.subscribe((newValue, oldValue) => {
    element.props[this.name] = newValue; // Bypasses set()
    this.changed(element, { ... });
});
```

This skips `parse()`, `convert()`, attribute reflection, and other side effects that `set()` handles.

### 4. Dual write in `set()` — `element.props` and `signal.value` can diverge

```js
element.props[this.name] = parsedValue;  // line ~242
// ... attribute reflection, etc ...
signal.value = parsedValue;              // line ~280
```

Between these two writes, any code triggered by attribute reflection that reads via the signal gets stale data. Also, `signal.value = parsedValue` may trigger computed recomputations that read `element.props`, creating temporal inconsistencies.

### 5. Unnecessary computed props for dependency-free function defaults

Old code only created a synthetic computed prop when `defaultDependencies.length > 0`. New code creates one for **every** function default, even `default: () => []`. This adds unnecessary signals, subscriptions, and entries to the `Props` map.

### 6. Documentation not updated

`src/plugins/props/README.md` still documents `dependencies`, `additionalDependencies`, and `defaultDependencies` which are all removed.

### 7. Missing tests

116 lines of dependency tests deleted, 0 signal-based equivalents added. Should test: computed prop updates, diamond dependencies (glitch-free), dynamic/conditional deps, `convert` with dependencies.

DmitrySharabin:
**Update (2026-04-27):** Items 1a, 1b, 2, 3, A, B, D fixed in #97 (merged). Item C remains as a follow-up. Strikethrough applied below; bodies kept for historical reference.

---

Testing this PR against consumer code surfaced three regressions vs `main` plus a few pre-existing rough edges that are worth knowing about. Mapping to the test-plan bullets in the PR summary.

## ~~Regressions on this PR — all fixed in #97~~

### ~~1. `propchange` events still fire for both plain and computed props — plain-prop regressions~~

#### ~~1a. Plain prop with explicit `defaultProp`, when the default prop changes~~

Pre-PR, `Prop.update(defaultProp)` fired `changed({source:"default"})` on the dependent whenever its `defaultProp` changed. That propagation is gone. The dependent prop has no `spec.get`/`spec.convert` so it's a plain `Signal` — auto-tracking never engages on the default prop. `prop.get` still falls through to `defaultProp.get(element)`, so reads are correct; only `propchange` is silently absent, which is the trappy shape.

```js
class Foo extends NudeElement {
  static props = {
    src:    { type: String, default: "initial" },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-1a", Foo);
let el = new Foo();
document.body.append(el);
el.src = "changed";
// expected: propchange for both "src" and "mirror"
// on this PR: "src = initial", "src = changed"; el.mirror reads "changed" but no event for it
```

Also affects initial propchange: `mirror` never fires on mount either.

#### ~~1b. Plain prop with a function default — `propchange` fires on the auto-generated name, not the declared name~~

The constructor unconditionally wraps every `typeof spec.default === "function"` into an auto-generated `defaultProp` ([Prop.js#L39–L49](https://github.com/nudeui/element/blob/signals-props/src/plugins/props/util/Prop.js#L39-L49)). The auto-generated prop is a `Computed` and fires its own `propchange` (on the auto-generated name, e.g. `defaultBar`). The declared prop (`bar`) is still a plain `Signal` and never fires `propchange` on its public name. Same underlying mechanic as 1a — consumers listen on the declared name and miss everything.

```js
class Foo extends NudeElement {
  static props = { bar: { default: () => 42 } };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-1b", Foo);
document.body.append(new Foo());
// expected: "bar = 42"
// on this PR: "defaultBar = 42" (twice — see 2), "bar" never fires
```

### ~~2. Initial `propchange` double-fires for every `Computed`-backed prop (spec.get and auto-generated `defaultProp`)~~

Two events fire on mount for the same computed prop, with the same value:

```js
class Foo extends NudeElement {
  static props = {
    base: { type: Number, default: 7 },
    derived: { get () { return this.base + 1; } },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-2", Foo);
document.body.append(new Foo());
// on this PR:
//   base = 7
//   derived = 8
//   derived = 8   ← duplicate
```

Same shape for auto-generated `defaultProp`s (`defaultBar` in 1b fires twice). Subsequent updates fire once — only the initial is doubled.

Appears to be two entry points both firing:
1. `Prop.initializeFor` hits `else if (element.props[name] === undefined && !this.defaultProp)` and fires `this.changed({source:"default"})` manually ([L206–L209](https://github.com/nudeui/element/blob/signals-props/src/plugins/props/util/Prop.js#L206-L209)).
2. The listener then reads the prop (`this[name]` inside `propChangedCallback`), which triggers the `Computed`'s first compute; its subscriber fires `#onComputedChange` → another `changed({source:"get"})`.

Plain props with only a literal default don't hit #2 (no `Computed`), so they fire once correctly — it's specifically the `Computed`-backed cases that regress.

### ~~3. Function defaults that reference other props update reactively — half-true (maps to test-plan bullet of the same name)~~

The auto-generated `defaultProp` recomputes on dep changes and fires `propchange`, but on the auto-generated name (`defaultDerived`). The declared prop (`derived`) stays silent. Same mechanic as 1b, just on update instead of init.

```js
class Foo extends NudeElement {
  static props = {
    base: { type: Number, default: 1 },
    derived: { type: Number, default () { return this.base * 10; } },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-3", Foo);
let el = new Foo();
document.body.append(el);
el.base = 2;
// expected: "derived = 20" after the last line
// on this PR: "defaultDerived = 20" fires, "derived" stays silent
```

---

## Pre-existing (also reproduces on `main` — flagging while we're in the area)

### ~~A. `spec.convert` + `default` + `reflect: true` — initial attribute reflection skipped~~

A convert prop with a default and `reflect: true` doesn't mirror the default value to the attribute on mount. The `propchange` fires with the resolved default, but `getAttribute` returns `null` until the prop is explicitly written.

```js
class Foo extends NudeElement {
  static props = {
    val: { convert (v) { return (v ?? 0) * 2; }, default: "hi", reflect: true },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name], "attr:", this.getAttribute(name)); }
}
customElements.define("x-foo-a", Foo);
document.body.append(new Foo());
// on both main and this PR: "val = hi attr: null"
```

### ~~B. Assigning the current (default-resolved) value fires `propchange` with `source:"property"`~~

Because the internal `Signal.value` is `undefined` until explicit set (defaults are only resolved lazily in `prop.get`), the equality check in `set()` compares the new value against `undefined`, not against the effective default — so `el.v = 0` fires even when `v` already reads `0`.

```js
class Foo extends NudeElement {
  propChangedCallback ({ detail }) { console.log("propchange source:", detail.source); }
}
customElements.define("x-foo-b", Foo);
let el = new Foo();
document.body.append(el);
el.v = 0;
// on both main and this PR:
//   propchange source: default    (initial)
//   propchange source: property   (from `el.v = 0` — unexpected, value didn't change)
```

### C. `propchange` events queued while disconnected are lost on reconnect

`firePropChangeEvent` queues events when `isConnected && initialized` is false. The queue is only drained inside `Props.initializeFor`, which runs once via `constructed` with `{once: true}`. Subsequent `disconnect → set → reconnect` cycles never flush the queue — those events disappear.

```js
class Foo extends NudeElement {
  propChangedCallback ({ name }) { console.log("propchange:", name, "connected:", this.isConnected); }
}
customElements.define("x-foo-c", Foo);
let el = new Foo();
document.body.append(el);   // "propchange: v connected: true"
el.remove();
el.v = 5;                    // queued silently
document.body.append(el);   // nothing — queue not drained
```

### ~~D. Defaults bypass `spec.convert`~~

When a prop has both `spec.convert` and a default (literal or function), the default value is returned as-is via `prop.get`'s fallback path — `spec.convert` is only applied to explicit writes. Authors who use `convert` to normalize values (e.g. `convert (c) { return c.to(this.space); }` with a default factory) may be surprised that the default escapes normalization.

---

## ~~Possible direction for (1) / (2) / (3)~~ — implemented in #97

The two-signal architecture already used for `spec.convert` (raw `Signal` + wrapping `Computed`) appears to generalize cleanly to cover `defaultProp` and literal/function defaults: every prop whose value comes from `convert` / `defaultProp` / `default` would use a `Computed` over a user-writable `rawSignal` that falls through to the right source. Writes target `rawSignal`; the Computed's subscriber would fire `propchange` uniformly on the declared name, once. A rough prototype along those lines made 1a / 1b / 3 fall out together, and the double-fire in 2 also goes away because the manual `initializeFor` fire is replaced by the Computed's initial-compute subscriber — one source of truth for when `propchange` fires. This would also make pre-existing (A) and (D) easier to address consistently, since the Computed would run `convert` on the default too.

DmitrySharabin:
**From Claude Code**:

Spinning Lea's review thread on `rawSignal.equals = (a, b) => this.equals(a, b);` ([#97 (comment)](https://github.com/nudeui/element/pull/97#discussion_r3149110630)) into a top-level discussion since it's a small design question worth resolving here before we keep adding signal-backed props.

The current code monkey-patches `equals` on every `Signal` / `Computed` instance after construction, in two spots ([Prop.js:155](https://github.com/nudeui/element/blob/signals-props/src/plugins/props/util/Prop.js#L155) and [Prop.js:188](https://github.com/nudeui/element/blob/signals-props/src/plugins/props/util/Prop.js#L188)). Three options:

### 1. Subclass

```js
// in Prop.js
class PropSignal extends Signal {
  #prop;
  constructor (prop, value) { super(value); this.#prop = prop; }
  equals (a, b) { return this.#prop.equals(a, b); }
}
class PropComputed extends Computed {
  #prop;
  constructor (prop, fn) { super(fn); this.#prop = prop; }
  equals (a, b) { return this.#prop.equals(a, b); }
}
```

- **Pro:** `equals` is part of the class contract, not a post-hoc patch. Signals.js stays minimal — no API change.
- **Con:** Two new subclasses for one method override. The `equals` method is duplicated across them (or gets factored into a mixin, which JS makes awkward).

### 2. `equals` constructor option

```js
// in signals.js
export class Signal extends EventTarget {
  constructor (value, { equals } = {}) {
    super();
    this.#value = value;
    if (equals) this.equals = equals;
  }
  ...
}
```

```js
// in Prop.js
let rawSignal = new Signal(undefined, { equals: (a, b) => this.equals(a, b) });
```

- **Pro:** No new classes. Both `Signal` and `Computed` accept it (Computed inherits via `super()`). The two `signal.equals = …` lines disappear — equality lives where the signal is born.
- **Con:** Tiny API addition to signals.js (one optional option).

### 3. Status quo — instance override

What's there today.

- **Pro:** Zero changes, works.
- **Con:** Reads as an afterthought; the equality contract is set up after construction, two locations to keep in sync.

### My read

I'd lean toward **(2)**. The "keep signals.js minimal" argument behind (1) is real, but a single optional constructor option is a smaller cost than two new subclasses + a duplicated method. (1) makes more sense if we expect more `Signal` customizations down the line — at that point a subclass that bundles them all is cleaner than a growing options bag.

Happy to implement whichever you and Lea land on.

DmitrySharabin:
## Test plan verification

Verified items 2–6 of the test plan via fresh browser evals against this branch with [#97](https://github.com/nudeui/element/pull/97) merged. Format matches the verification comment on the child PR — class definition, actions, `expected:` / `actual:` inline. Item 1 (`npm test`) is unverified — suite is broken on this branch. Item 5's setAttribute-after-mount edge case is a pre-existing setup quirk filed as #98.

### 2. Computed props update when dependencies change

```js
class Foo extends NudeElement {
  static props = {
    base:    { type: Number, default: 1 },
    derived: { get () { return this.base * 2; } },
  };
}
customElements.define("x-foo-tp2", Foo);
let el = new Foo();
document.body.append(el);
// expected: el.derived === 2 on mount; updates as base changes
// actual:   2

el.base = 5;
// expected: el.derived === 10
// actual:   10

el.base = 10;
// expected: el.derived === 20
// actual:   20
```

### 3. Dynamic/conditional deps

```js
class Foo extends NudeElement {
  static props = {
    cond: { default: true },
    a:    { type: Number, default: 100 },
    b:    { type: Number, default: 200 },
    out:  { get () { return this.cond ? this.a : this.b; } },
  };
  propChangedCallback ({ name }) { if (name === "out") console.log("out =", this.out); }
}
customElements.define("x-foo-tp3", Foo);
let el = new Foo();
document.body.append(el);
// expected: el.out === 100 (cond=true → reads a)
// actual:   100

el.cond = false;
// expected: out re-tracks deps and now reads b
// actual:   "out = 200"

el.a = 999;
// expected: a is no longer a dep when cond=false; out should NOT change
// actual:   el.out still 200; no propchange for out

el.b = 777;
// expected: b is the dep now; out updates
// actual:   "out = 777"
```

### 4. `propchange` events fire for both plain and computed props

```js
class Foo extends NudeElement {
  static props = {
    plain:     { type: Number, default: 1 },                              // plain Signal-backed
    computed:  { get () { return this.plain + 10; } },                     // spec.get Computed
    fnDefault: { type: Number, default () { return this.plain * 100; } }, // function-default Computed
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-tp4", Foo);
let el = new Foo();
document.body.append(el);
// expected: all three fire on mount
// actual:   "plain = 1", "computed = 11", "fnDefault = 100"

el.plain = 5;
// expected: all three fire on update
// actual:   "plain = 5", "computed = 15", "fnDefault = 500"
```

### 5. Attribute reflection still works for plain props

```js
class Foo extends NudeElement {
}
customElements.define("x-foo-tp5", Foo);

// 5a — prop → attribute (write to property reflects to attribute)
let el = new Foo();
document.body.append(el);
el.v = 42;
// expected: el.getAttribute("v") === "42"
// actual:   "42"

// 5b — pre-set attribute on mount → property is parsed
let el2 = document.createElement("x-foo-tp5");
el2.setAttribute("v", "55");
document.body.append(el2);
// expected: el2.v === 55
// actual:   55
```

`setAttribute()` *after* mount does not update the property — pre-existing issue, see #98.

### 6. Function defaults that reference other props update reactively

```js
class Foo extends NudeElement {
  static props = {
    base:    { type: Number, default: 3 },
    derived: { type: Number, default () { return this.base * 100; } },
  };
  propChangedCallback ({ name }) { if (name === "derived") console.log("derived =", this.derived); }
}
customElements.define("x-foo-tp6", Foo);
let el = new Foo();
document.body.append(el);
// expected: el.derived === 300 on mount
// actual:   300

el.base = 7;
// expected: derived recomputes via function default; fires on declared name
// actual:   "derived = 700"
```

DmitrySharabin (approved):

PR #92, opened 2026-03-20
Add bundler compatibility to styles plugin

DmitrySharabin:
## Summary

`getStyle()` normalizes any style input into an options object with a ready-to-adopt `css` property (`CSSStyleSheet` or `Promise<CSSStyleSheet>`). This enables bundler-compatible patterns like `new URL("./foo.css", import.meta.url)` and CSS Import Attributes.

### Changes

- **`get-style.js`**: Single `getStyle(value, baseUrl, defaults)` function handling all input types:
  - `CSSStyleSheet` — returned as-is
  - `string` / `URL` — resolved via `cssToSheet(cachedFetch(fullUrl))`
  - Options dict — URL coercion, `css ??=` (explicit css wins over fetch), CSS string → `CSSStyleSheet`
  - `Promise` — resolves, unwraps ESM `.default` if present, recurses through `getStyle`
  - Spread order `{ ...defaults, ... }` ensures explicit values win
- **`base.js`**: `defineStyles` simplified to one `getStyle()` call per entry; deduplication uses `options.fullUrl` directly
- **`util.js`**: Barrel exports `getStyle`
- **Tests**: 10 declarative hTest cases covering all branches. `run()` is a one-liner pass-through; Promise tests verify `css` is a Promise resolving to `CSSStyleSheet`

Partially addresses #89.


src/plugins/styles/base.js
```diff
@@ -64,7 +64,23 @@ const providesStatic = {
 		const baseUrl = this.url ?? defaultBaseURL;
 
 		for (let options of def) {
-			if (typeof options === "string") {
+			if (options instanceof URL) {
```
LeaVerou: Not all `URL` instances are due to bundlers, these may come up naturally for a number of other reasons. 
This seems to conflate two things:
- Styles could be either a URL reference or an options dictionary
- Style URL could be either a string or a `URL` object

There are 4 combinations of conditions here, and this only handles 3 of them. A natural way to handle this without any increase in complexity is:
- Check if `options` is a URL → if so, rewrite to `{ url: options }`
- Check if `options.url` is a `URL` object → if so, do `url = url.href` (if that's even needed, most APIs that handle string URLs also handle `URL` instances since they naturally get coerced to `url.href`).

LeaVerou (changes requested):
I left some comments, but also this is getting to the point where it needs a utility function (e.g. `getStyle()`), rather than doing everything in `base.js`.

LeaVerou (approved):
Nit: it may be better to use `Object.create()` to encode the `defaults` as a parent instead of subsuming them in the returned object. But that's not a blocker.

PR #96, opened 2026-04-07
Replace local extensibility core with xtensible package

LeaVerou:
## Summary

- Replace the local plugin system (`extensible.js`, `hooks.js`, `symbols.js`, `plugins.js`, `base.js`, `plugins-property/`) with the extracted `xtensible` package
- Update all plugin files (~25) to import from `xtensible` and `xtensible/util` instead of local paths
- Use `$hook`, `hooksCommon`, and `api` from `xtensible/plugins` instead of the old `base` and `plugins-property` plugins
- Update `factory.js` to match the pattern already established in `element.js`
- Delete 5 local files now provided by `xtensible` (net -457 lines)

## Test plan

- [x] `npm test` — same 57/71 pass as clean main (11 pre-existing failures, 0 regressions)
- [x] Manual smoke test with a consuming component (slider example)
- [x] Verify no remaining imports from deleted local files


DmitrySharabin (approved):
I pushed a commit to get rid of stale imports and fixed one import (even though it was in a currently unused file). The changes in the PR LGTM and seem to work fine. However, we need to make sure the https://github.com/LeaVerou/xtensible/issues/1 bug is fixed and the new version of `xtensible` is published before merging it.

**UPDATE:** The issue is fixed.

PR #97, opened 2026-04-27
Unify default handling under Computed-backed pattern

DmitrySharabin:
## Summary

Follow-up to #91. Fixes the bugs in the review and unifies default handling under the Computed-backed pattern.

### Bugs fixed (review scenarios from #91)

- 1a: `defaultProp` source-change propagation
- 1b: function default fires on declared name (not synthetic `defaultBar`)
- 2: no double-fire on mount for Computed-backed props
- 3: function defaults reactive on declared name
- Sever-on-write, restore-on-undefined for `defaultProp`
- Parse-on-default coercion: function defaults' return values now coerced via `parse()` (and `convert`, if specified)

### Architectural change

`getSignal` collapses to two default-handling shapes:

- Computed props (`spec.get`): plain Computed.
- **Any prop with a default OR a convert**: raw Signal + Computed wrapper. The Computed body resolves defaults, applies parse, and runs convert in one place.

Synthetic `default<Name>` props are gone from the registry.

### Pre-existing items also fixed (incidentally)

- **A**: defaults with `reflect: true` now reflect to the attribute on mount — covers both `convert+default+reflect` (the case in the original review) and plain `default+reflect`.
- **B**: assigning the current default-resolved value no longer fires a redundant propchange.
- **D**: defaults now pass through `spec.convert`.

### Behavior delta worth flagging

`change.source` for user writes on Computed-backed props is now `"default"` / `"convert"` (the construction-time path), not `"property"` / `"attribute"`. Pre-#91, function-default props were plain Signal and reported `"property"`/`"attribute"`. This unifies their behavior with `spec.convert`, which already reported `"convert"` regardless of trigger. No source consumers grep'd in tree.

### Out of scope

Item C from the #91 review (queued events lost on disconnect/reconnect) — not touched by this PR.


DmitrySharabin:
## Verification

Ran each of the scenarios from the original review against this branch. Each block sets up a fresh element, runs the actions, and shows the resulting events / state. Format mirrors the original review comment on #91.

### Bugs from the #91 review

#### 1a. `defaultProp` source-change propagation

```js
class Foo extends NudeElement {
  static props = {
    src:    { type: String, default: "initial" },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-1a", Foo);
let el = new Foo();
document.body.append(el);
el.src = "changed";
// expected: propchange for both "src" and "mirror"
// actual:   "src = initial", "mirror = initial", "src = changed", "mirror = changed"
```

#### 1b. Function default fires on declared name (not synthetic `defaultBar`)

```js
class Foo extends NudeElement {
  static props = { bar: { default: () => 42 } };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-1b", Foo);
document.body.append(new Foo());
// expected: "bar = 42"
// actual:   "bar = 42"
```

#### 2. No double-fire on mount for Computed-backed props

```js
class Foo extends NudeElement {
  static props = {
    base: { type: Number, default: 7 },
    derived: { get () { return this.base + 1; } },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-2", Foo);
document.body.append(new Foo());
// expected: "base = 7", "derived = 8" (each fires once on mount)
// actual:   "base = 7", "derived = 8"
```

#### 3. Function defaults reactive on declared name

```js
class Foo extends NudeElement {
  static props = {
    base: { type: Number, default: 1 },
    derived: { type: Number, default () { return this.base * 10; } },
  };
  propChangedCallback ({ name }) { console.log(name, "=", this[name]); }
}
customElements.define("x-foo-3", Foo);
let el = new Foo();
document.body.append(el);
el.base = 2;
// expected: "derived = 20" after the last line (fires on declared name)
// actual:   "base = 1", "derived = 10", "base = 2", "derived = 20"
```

#### Sever-on-write, restore-on-undefined for `defaultProp`

```js
class Foo extends NudeElement {
  static props = {
    src:    { type: String, default: "initial" },
  };
}
customElements.define("x-foo-sever", Foo);
let el = new Foo();
document.body.append(el);

el.mirror = "explicit"; // sever
// expected: el.mirror === "explicit"
// actual:   "explicit"

el.src = "after-sever";
// expected: el.mirror still "explicit" (src no longer flows through)
// actual:   "explicit"

el.mirror = undefined; // restore
// expected: el.mirror === "after-sever"
// actual:   "after-sever"
```

#### Parse-on-default coercion

```js
class Foo extends NudeElement {
  static props = {
    n: { type: Number, default: () => "42" },
  };
}
customElements.define("x-foo-parse", Foo);
let el = new Foo();
document.body.append(el);
// expected: el.n === 42 (number, coerced via parse)
// actual:   42, typeof "number"
```

### Pre-existing items also verified

#### A. Defaults with `reflect: true` reflect on mount

```js
class Foo extends NudeElement {
  static props = {
    val:   { type: Number, default: 5, convert (v) { return v * 2; }, reflect: true },
    plain: { type: Number, default: 7, reflect: true },
  };
}
customElements.define("x-foo-a", Foo);
let el = new Foo();
document.body.append(el);
// expected: el.getAttribute("val") === "10", el.getAttribute("plain") === "7"
// actual:   "10", "7"
```

#### B. No redundant propchange when assigning the current default-resolved value

```js
class Foo extends NudeElement {
  propChangedCallback ({ detail }) { console.log("source:", detail.source); }
}
customElements.define("x-foo-b", Foo);
let el = new Foo();
document.body.append(el);
el.v = 0;
// expected: only the initial mount event fires (assigning current value is a no-op)
// actual:   "source: default" — nothing else
```

#### D. Defaults pass through `spec.convert`

```js
class Foo extends NudeElement {
  static props = {
    n: { type: Number, default: 5, convert (v) { return v * 2; } },
  };
}
customElements.define("x-foo-d", Foo);
let el = new Foo();
document.body.append(el);
// expected: el.n === 10 (default 5, convert *2)
// actual:   10
```

#### Plain literal default sever/restore via `el.v = undefined`

```js
class Foo extends NudeElement {
}
customElements.define("x-foo-case4", Foo);
let el = new Foo();
document.body.append(el);

el.v = 100;
// expected: el.v === 100
// actual:   100

el.v = undefined;
// expected: el.v === 0 (default restored, not wiped to undefined)
// actual:   0
```

#### `null` preserved on Computed-backed props with default

```js
class Foo extends NudeElement {
  static props = {
    base: { type: Number, default: 7 },
    v:    { type: Number, default () { return this.base * 10; } },
  };
}
customElements.define("x-foo-case5", Foo);
let el = new Foo();
document.body.append(el);

el.v = null;
// expected: el.v === null (null preserved, not clobbered by default)
// actual:   null

el.v = undefined;
// expected: el.v === 70 (default restored)
// actual:   70
```

Item C from the original review (queued events lost on disconnect/reconnect) is not addressed by this PR — left as a follow-up.

src/plugins/props/util/Prop.js
```diff
@@ -158,22 +148,38 @@ let Self = class Prop {
...
 				});
 			}
-			else if (this.spec.convert) {
-				// Two-signal pattern: raw holds pre-convert value,
-				// Computed applies convert and auto-tracks its dependencies
+			else if (this.spec.convert || this.spec.default !== undefined) {
+				// Raw Signal holds the user-set value; Computed wraps it to apply
+				// convert and/or fall through to the default. Auto-tracks deps.
 				let rawSignal = new Signal(undefined);
 				rawSignal.equals = (a, b) => this.equals(a, b);
```
LeaVerou, line 155: Should we subclass instead of overridding? Or have an `equals` constructor option? Also creating it here seems a bit weird?

LeaVerou (approved):

src/plugins/props/util/Prop.js
DmitrySharabin: Probably one of these is a better approach. I'd leave it as a follow-up improvement to the parent PR, not this one.

PR #102, opened 2026-04-29
Batch propchange events on a microtask drain

### LINKED ISSUE #51: Bunch up property changes
`propChangedCallback` gets called too frequently, and it's hard to get a good picture of what changed from just a single property. We should wait until changes settle and collect them meanwhile.

Lit does something similar too. We could probably even use the same signature.

### LINKED ISSUE #100: `propchange` events queued while disconnected are not dispatched on reconnect
## Symptom

For a prop on an already-mounted element, mutating the prop while the element is disconnected drops the resulting `propchange` event — it is queued internally but never dispatched when the element reconnects.

```js
class C extends NudeElement {
}

let el = new C();
document.body.append(el);
el.addEventListener("propchange", e => console.log(e.name));

el.remove();             // disconnect
el.v = 5;                // mutation lands in the queueing branch
await Promise.resolve(); // let the Computed microtask flush while disconnected
document.body.append(el); // reconnect

// expected: console logged "v"
// actual:   nothing — queued event never drained
```

## Root cause

`firePropChangeEvent` enters its queueing branch when the element is disconnected ([plugins/props/util/Props.js#L74-L81](https://github.com/nudeui/element/blob/main/src/plugins/props/util/Props.js#L74-L81)):

```js
if (element.isConnected && eventProps.prop.initialized) {
    element.dispatchEvent?.(event);
}
else {
    let queue = this.eventDispatchQueue.get(element) ?? [];
    queue.push(event);
    this.eventDispatchQueue.set(element, queue);
}
```

The queue is drained exactly once, inside `Props.initializeFor` ([plugins/props/util/Props.js#L100-L109](https://github.com/nudeui/element/blob/main/src/plugins/props/util/Props.js#L100-L109)), and `initializeFor` is invoked only from the `constructed` lifecycle hook ([plugins/props/index.js#L40-L42](https://github.com/nudeui/element/blob/main/src/plugins/props/index.js#L40-L42)) — i.e. once per element, at construction time. The plugin registers no `connected` hook, so there is nothing to drain the queue when `isConnected` later flips back to `true`. Events that landed in the queueing branch after construction are permanently stuck in the `eventDispatchQueue` `WeakMap`.

## Reproducible on

`main` and the `props-runtime-tests` branch. Surfaced as a skipped JS-first test in [`test/Props.js`](https://github.com/nudeui/element/blob/props-runtime-tests/test/Props.js) ("Queued propchange events drain on reconnect"). Pre-existing — not introduced by #79 / #91 / #97; the test was added in #99 specifically to track this case.

## Suggested fix direction

Split a `drainEventQueue(element)` helper out of `initializeFor` and call it from a `connected` lifecycle hook (or `connectedCallback`, whichever fits the plugin contract — see how `attributeChangedCallback` is wired in [plugins/props/index.js#L10-L14](https://github.com/nudeui/element/blob/main/src/plugins/props/index.js#L10-L14)). Calling `initializeFor` itself on every reconnect would over-fire — it walks observed attributes and re-runs `prop.initializeFor` for every prop, which is mount-time work, not reconnect-time work.

Once fixed, un-skip the corresponding test in `test/Props.js`.

DmitrySharabin:
## Summary

`propchange` handlers were observing intermediate state during multi-prop synchronous writes. Microtask-batch dispatch so handlers always see settled state, plus a `connected` lifecycle hook that drains queued events on (re)connect.

Closes #51, closes #100.

> Stacked on #108. Base auto-updates to `main` after #108 merges.

## Why

```js
class MyComponent extends NudeElement {
  static props = { color: {…}, space: {…} };

  propChangedCallback ({ name }) {
    // assumes both are in sync
    update(this.color, this.space);
  }
}

// caller:
el.color = newColor; // handler runs — new color, OLD space
el.space = newSpace; // handler runs — finally settled
```

Microtask coalescing collapses sync writes to one dispatch per prop with the settled value, and unifies timing across plain Signal and Computed-backed props.

## API

| Granularity      | Event         | Auto-wired callback          |
| ---------------- | ------------- | ---------------------------- |
| Per prop         | `propchange`  | `propChangedCallback(event)` |
| Per drain (bulk) | `propschange` | `updated(event)`             |

```js
class MyElement extends NudeElement {
  static props = { foo: { /* … */ }, bar: { /* … */ } };

  propChangedCallback (event) {
    console.log(event.name, event.detail.value);
  }

  updated (event) {
    for (let [name, oldValue] of event.changedProps) {
      console.log(name, oldValue, "→", this[name]);
    }
  }
}
```

`propschange` fires once at the end of every drain, after every per-prop `propchange`. `event.changedProps` is `Map<name, oldValue>`; read the current value via `this[name]`.

## Implementation

- **Coalescing** — sync writes to the same prop collapse to one event per drain. Latest `value` / `source` win; `oldValue` and `oldAttributeValue` pin to the first write so the payload spans the full first→last delta.
- **Round-trip filter** — drain-time `prop.equals` skips coalesced no-ops (`el.foo = X; el.foo = oldValue`).
- **Reflection** is synchronous in `Prop.set` and `#onComputedChange` — not deferred to the drain.
- **`connected(element)`** drains the per-element dispatch queue on (re)connect (closes #100).
- **`updated()` listener uses a stable method reference**, so subclass re-attachment dedupes per the DOM spec.

## Test plan

- [x] `npm test -- --ci` — 99/100 PASS, 1 pre-existing skip, 0 FAIL

## Follow-ups

- **#104** — `static props` not inherited from a parent class. Latent; no consumer hits it today.
- **#98** — `setAttribute` after mount silently fails. Once fixed, the `!isConnected` guard in `Props#attributeChanged` comes out.

### DIFF
--- a/src/plugins/events/propchange.js
+++ b/src/plugins/events/propchange.js
@@ -41,7 +41,10 @@ const hooks = {
 	},
 
 	first_connected () {
-		// Often propchange events have already fired by the time the event handlers are added
+		// Don't remove: re-fires initial propchange for on*= attribute handlers,
+		// which onprops attaches *after* the initial dispatch. Without this,
+		// shortcut handlers declared in HTML never see the initial event.
+		// Pre-connect imperative listeners receive the event twice.
 		for (let eventName in this.constructor[propchange]) {
 			let propName = this.constructor[propchange][eventName];
 			let value = this[propName];
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -30,8 +30,18 @@ const hooks = {
 	},
 
 	constructor () {
-		if (this.propChangedCallback && this.constructor[props]) {
+		if (!this.constructor[props]) {
+			return;
+		}
+
+		// Per-prop callback: auto-wire to propchange events.
+		if (this.propChangedCallback) {
 			this.addEventListener("propchange", this.propChangedCallback);
+		}
+
+		// Bulk callback: auto-wire to the propschange event.
+		if (this.updated) {
+			this.addEventListener("propschange", this.updated);
 		}
 	},
 
@@ -39,6 +49,10 @@ const hooks = {
 
 	constructed () {
 		this.constructor[props].initializeFor(this);
+	},
+
+	connected () {
+		this.constructor[props].connected(this);
 	},
 };
 
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -106,8 +106,8 @@ let Self = class Prop {
 		this.changed(element, {
 			element,
 			source,
-			parsedValue: newValue,
-			oldInternalValue: oldValue,
+			value: newValue,
+			oldValue,
 		});
 	}
 
@@ -217,9 +217,7 @@ let Self = class Prop {
 				// Force first compute so the subscriber emits the initial propchange
 				signal.value;
 			}
-			else {
-				this.changed(element, { source: "default", element });
-			}
+			// Plain Signals start at undefined: nothing to fire about at mount.
 		}
 
 		this.#initialized = true;
@@ -256,14 +254,13 @@ let Self = class Prop {
 		return signal.value;
 	}
 
-	set (element, value, { source, name, oldValue } = {}) {
+	set (element, value, { source, name, oldAttributeValue } = {}) {
 		let signal = this.getSignal(element);
 		let rawSignal = this.#rawSignals.get(element);
 
 		// For Computed-backed props, compare against the raw user-set value
-		let oldInternalValue = (rawSignal ?? signal).value;
+		let oldValue = (rawSignal ?? signal).value;
 
-		let attributeName = name;
 		let parsedValue;
 
 		try {
@@ -285,7 +282,7 @@ let Self = class Prop {
 			parsedValue = undefined;
 		}
 
-		if (this.equals(parsedValue, oldInternalValue)) {
+		if (this.equals(parsedValue, oldValue)) {
 			return;
 		}
 
@@ -302,10 +299,8 @@ let Self = class Prop {
 			let change = {
 				element,
 				source,
-				value,
-				parsedValue,
-				oldInternalValue,
-				attributeName: name,
+				value: parsedValue,
+				oldValue,
 			};
 
 			if (source === "property") {
@@ -315,7 +310,7 @@ let Self = class Prop {
 					let oldAttributeValue = element.getAttribute(attributeName);
 
 					if (oldAttributeValue !== attributeValue) {
-						element.ignoredAttributes.add(this.toAttribute);
+						element.ignoredAttributes.add(attributeName);
 
 						Object.assign(change, { attributeName, attributeValue, oldAttributeValue });
 						this.applyChange(element, { ...change, source: "attribute" });
@@ -326,9 +321,9 @@ let Self = class Prop {
 			}
 			else if (source === "attribute") {
 				Object.assign(change, {
-					attributeName,
+					attributeName: name,
 					attributeValue: value,
-					oldAttributeValue: oldValue,
+					oldAttributeValue,
 				});
 			}
 
--- a/src/plugins/props/util/Props.js
+++ b/src/plugins/props/util/Props.js
@@ -1,7 +1,31 @@
 import Prop from "./Prop.js";
 import PropChangeEvent from "./PropChangeEvent.js";
+import PropsChangeEvent from "./PropsChangeEvent.js";
 
 export default class Props extends Map {
+	/**
+	 * Per-element coalescing buffer. Sync writes to the same prop overwrite the entry's
+	 * `value`/`source` (latest wins) while pinning `oldValue`/`oldAttributeValue` to the
+	 * first write — so the drained payload spans the full first→last delta. Drained and
+	 * cleared in `#drainFor`.
+	 * @type {WeakMap<HTMLElement, Map<string, {name: string, prop: Prop, detail: object}>>}
+	 */
+	#eventDispatchQueue = new WeakMap();
+
+	/**
+	 * Elements with queued events awaiting drain. Snapshotted and cleared at the top of
+	 * `#drain` so re-entrant writes (from handlers) accumulate for the *next* drain.
+	 * @type {Set<HTMLElement>}
+	 */
+	#pendingElements = new Set();
+
+	/**
+	 * Microtask-schedule guard. Set when a drain is queued, cleared when it runs —
+	 * collapses many `propChanged` calls in the same tick into one `queueMicrotask`.
+	 * @type {boolean}
+	 */
+	#drainScheduled = false;
+
 	/**
 	 *
 	 * @param {HTMLElement} Class The class to define props for
@@ -46,39 +70,103 @@ export default class Props extends Map {
 		let propsFromAttribute = [...this.values()].filter(spec => spec.fromAttribute === name);
 
 		for (let prop of propsFromAttribute) {
-			prop.set(element, element.getAttribute(name), { source: "attribute", name, oldValue });
-		}
-	}
-
-	eventDispatchQueue = new WeakMap();
-
-	/**
-	 * Called when a prop value changes. Fires propchange events.
-	 * Dependency propagation is handled automatically by signals.
-	 */
-	propChanged (element, prop, change) {
-		// Fire propchange event
-		let eventNames = ["propchange", ...(prop.eventNames ?? [])];
-		for (let eventName of eventNames) {
-			this.firePropChangeEvent(element, eventName, {
-				name: prop.name,
-				prop,
-				detail: change,
+			prop.set(element, element.getAttribute(name), {
+				source: "attribute",
+				name,
+				oldAttributeValue: oldValue,
 			});
 		}
 	}
 
-	firePropChangeEvent (element, eventName, eventProps) {
-		let event = new PropChangeEvent(eventName, eventProps);
+	/**
+	 * Called from Prop#changed when a value settles. Coalesces into the
+	 * dispatch queue; dispatch happens in #drainFor on the next microtask.
+	 */
+	propChanged (element, prop, change) {
+		let map = this.#eventDispatchQueue.get(element);
+		if (!map) {
+			map = new Map();
+			this.#eventDispatchQueue.set(element, map);
+		}
 
-		if (element.isConnected && eventProps.prop.initialized) {
-			element.dispatchEvent?.(event);
+		let existing = map.get(prop.name);
+		if (existing) {
+			// Coalesce: latest value/source wins, but old values stay pinned
+			// to the first write so the payload spans the full first→last delta.
+			let { oldValue, oldAttributeValue } = existing.detail;
+			Object.assign(existing.detail, change, { oldValue, oldAttributeValue });
 		}
 		else {
-			let queue = this.eventDispatchQueue.get(element) ?? [];
-			queue.push(event);
-			this.eventDispatchQueue.set(element, queue);
+			map.set(prop.name, { name: prop.name, prop, detail: { ...change } });
 		}
+
+		this.#pendingElements.add(element);
+		this.#scheduleDrain();
+	}
+
+	#scheduleDrain () {
+		if (this.#drainScheduled) {
+			return;
+		}
+
+		this.#drainScheduled = true;
+		queueMicrotask(() => {
+			this.#drainScheduled = false;
+			this.#drain();
+		});
+	}
+
+	#drain () {
+		// Snapshot and clear: events queued by handlers (incl. on other
+		// elements) run on the next microtask, not in this drain.
+		let elements = [...this.#pendingElements];
+		this.#pendingElements.clear();
+
+		for (let element of elements) {
+			this.#drainFor(element);
+		}
+	}
+
+	#drainFor (element) {
+		if (!element.isConnected) {
+			// Queue stays intact; `connected()` drains it on (re)connect.
+			return;
+		}
+
+		let map = this.#eventDispatchQueue.get(element);
+		if (!map) {
+			return;
+		}
+
+		// Detach the queue before dispatch: re-entrant writes from event
+		// handlers must accumulate for the next drain, not this one.
+		let entries = [...map];
+		this.#eventDispatchQueue.delete(element);
+
+		let changedProps = new Map();
+		for (let [, payload] of entries) {
+			// Plain Signals don't dedupe coalesced round-trips on their own;
+			// mirror Signal equality here.
+			let { prop, detail } = payload;
+			if (prop.equals(detail.value, detail.oldValue)) {
+				continue;
+			}
+
+			changedProps.set(prop.name, detail.oldValue);
+
+			// EventTarget isolates listener throws — siblings stay safe without try/catch.
+			for (let name of ["propchange", ...(prop.eventNames ?? [])]) {
+				element.dispatchEvent(new PropChangeEvent(name, payload));
+			}
+		}
+
+		if (changedProps.size > 0) {
+			element.dispatchEvent(new PropsChangeEvent("propschange", { changedProps }));
+		}
+	}
+
+	connected (element) {
+		this.#drainFor(element);
 	}
 
 	initializeFor (element) {
@@ -97,15 +185,7 @@ export default class Props extends Map {
 			prop.initializeFor(element);
 		}
 
-		// Dispatch any events that were queued
-		let queue = this.eventDispatchQueue.get(element);
-
-		if (queue) {
-			for (let event of queue) {
-				element.dispatchEvent?.(event);
-			}
-
-			this.eventDispatchQueue.delete(element);
-		}
+		// Drain synchronously so callers see initial state without waiting for the microtask.
+		this.#drainFor(element);
 	}
 }
--- a/src/plugins/props/util/PropsChangeEvent.js
+++ b/src/plugins/props/util/PropsChangeEvent.js
@@ -1,0 +1,7 @@
+export default class PropsChangeEvent extends CustomEvent {
+	constructor (type, { changedProps, ...options } = {}) {
+		super(type, options);
+
+		this.changedProps = changedProps;
+	}
+}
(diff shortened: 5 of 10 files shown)


LeaVerou:
What if we keep `propsChangedCallback()` low-level like it is now and introduce a new `updated()` (and corresponding event) with the batched map?

src/plugins/props/index.js
```diff
@@ -29,17 +29,32 @@ const hooks = {
...
+		if (this.propChangedCallback) {
 			this.addEventListener("propchange", this.propChangedCallback);
 		}
-	},
 
-	first_constructor_static,
+		// Bulk callback: auto-wire to the propsupdate event, unwrap so it
+		// receives the changed-props Map directly.
+		if (this.updated) {
+			this.addEventListener("propsupdate", e => this.updated(e.changedProps));
```
LeaVerou: The way `this.propChangedCallback()` is attached, attaching it multiple times (e.g. due to subclasses) is ok. The way this is attached, it will fire multiple times. Is that a problem? I suspect it is but haven't thought about it deeply.

LeaVerou:
Haven't look at the code yet in detail, but on a high level,

image

Perhaps this is warranted, but at first glance, it feels disproportionate with the magnitude of the change which makes me worry that Claude may have overengineered this, as it often does.

The way I was thinking about it, we'd fire propchange as usual (sync, per property), but also have a debounced `propschange` (or `propsupdate`) that fires a certain delay after the *last* prop change (i.e. debounced). 

This could live in the Props controller for a given element (perhaps make an `ElementProps` class instead of having `method(element)` on `Props`). It would listen to propchange, collect events, drain XXms after the last propchange OR wait until the element gets connected. For disconnected elements, I'd expect the low-level propchange to still fire, but not the higher level propschange. But that's not a strong opinion.

It seems that should be a fairly small and straightforward change, not ...whatever this is that Claude did. But perhaps I'm missing some factors that necessitated this complexity?

---

How does it behave with deep class hierarchies? Does updated() get fired with all props or just the class's own props? What behavior do we want? Would be useful to look at some of our own deep hierarchies and see what would be most useful.

-----

> `console.log(name, payload.detail.oldInternalValue, "→", payload.detail.parsedValue);`

Why do we need `detail`? What does `payload` give us by itself? Also, do we really need a whole object? Lit just does prop → old value. I can see how having both internal and parsed value can be useful. Do we need anything else? If not, I wonder if we should just have two simple maps. Or, we just provide the internal values, and people can always call `parse()` themselves. 🤔 

-----

For later: A lot of this seems useful for any class, the only part that's specific to custom elements is the attribute-property reflection. I wonder if at some point later we could extract the rest as a separate helper for reactive classes.

DmitrySharabin:
> Haven't look at the code yet in detail, but on a high level,
> image
> 
> Perhaps this is warranted, but at first glance, it feels disproportionate with the magnitude of the change...

The real picture is not that bad, actually. The visible `+824` is 67% tests and 7% README. The actual production-code change is `+216` / `−88` (net `+128`), and `Prop.js` inside that is a net `−44` (reflection extraction is straight cleanup).

Initially, the actual driver of the PR wasn't #51; it was that `propchange` handlers observe intermediate states during multi-prop writes. For example,

```js
class MyComponent extends NudeElement {
  static props = { color: {...}, space: {...} };
  propChangedCallback ({ name }) {
    // reads BOTH; assumes they're consistent
    update(this.color, this.space);
  }
}

// caller does:
el.color = newColor;  // handler runs — sees new color, OLD space
el.space = newSpace;  // handler runs — finally settled
```
Batching (and coalescing) is what fixes it. That's why it's also in this PR (and closes #51, but it's more of a consequence).

We also unify timing: today, plain Signal props fire `propchange` sync, Computed-backed props fire asynchronously (next microtask). With the PR, both fire on the drain. As a result, we have one consistent timing model.

> How does it behave with deep class hierarchies? Does updated() get fired with all props or just the class's own props?

Based on the investigation, I filed an issue (#104).

Speaking of the changes in this PR, each class with static props gets its own Props instance with its own `#eventDispatchQueue`, `#pendingElements`, `#drainScheduled`. Independent drains, independent dispatches.

> Why do we need `detail`? What does `payload` give us by itself? Also, do we really need a whole object? Lit just does prop → old value. I can see how having both internal and parsed value can be useful. Do we need anything else? If not, I wonder if we should just have two simple maps. Or, we just provide the internal values, and people can always call `parse()` themselves. 🤔

Agreed. A full bag is overkill. If we don't need all, I'd do it like Lit does. If a consumer needs source semantics or attribute round-trips, etc., that's what `propChangedCallback` already covers. Per-prop event shape stays as-is.

In the codebase, internal and parsed values are effectively the same thing. The signal stores the parsed/converted value; “internal” was just our naming for it. There's no separate raw form to expose. Tracking both was over-engineering.

If you like, I can split the PR into 2 or 3 separate ones. For example,
1. `first_connected` re-fire `detail` population fix (independent quality fix).
2. Settled-state coalescing + reconnect drain. Drops some over-engineered bits.
3. Bulk callback (`updated`/`propsupdate`).

Alternatively, we can have everything in one shot. 😅

src/plugins/props/index.js
DmitrySharabin: Thank you for pointing this out. You are right. Fixed.

DmitrySharabin:
Real code changes are not that drastic, I suppose:
image

`+152` / `-62` (net `+90`)

LeaVerou:
> Initially, the actual driver of the PR wasn't https://github.com/nudeui/element/issues/51; it was that propchange handlers observe intermediate states during multi-prop writes.

But that's exactly what #51 is about!

> Agreed. A full bag is overkill. If we don't need all, I'd do it like Lit does. If a consumer needs source semantics or attribute round-trips, etc., that's what propChangedCallback already covers. Per-prop event shape stays as-is.

They do separate things though. I can see use cases where you might want all the info, but they're not the majority use case. But since we already have the detailed objects, it's also cheap to construct that map *as well*. So I'm leaning towards a signature like `updated(simpleMap, mapWithAllTheDetails)`, then consumers with weird use cases can always use the second argument, but most consumers will be fine with the first.

OTOH we can always ship with the first argument only, and add the second if use cases come up that need it. Yes, let's do that actually. Simplicity first!

> In the codebase, internal and parsed values are effectively the same thing. The signal stores the parsed/converted value; “internal” was just our naming for it. There's no separate raw form to expose. Tracking both was over-engineering.

I thought there were cases where we stored one value and the getter returned another but I could be misremembering.

> If you like, I can split the PR into 2 or 3 separate ones. For example,

I think splitting 1 into a separate PR makes sense, if it's an independent fix. 2 and 3 seem more intertwined.

I still think `propschange` is a better event name than `propsupdate`. I can see the argument that it's only 1 character and can be mistyped, but the other one makes me wonder what the difference between `change` and `update` is.

Thanks re:subclasses. I agree we can tackle that separately.

src/plugins/events/propchange.js
```diff
@@ -41,7 +41,10 @@ const hooks = {
 	},
 
 	first_connected () {
-		// Often propchange events have already fired by the time the event handlers are added
+		// `onprops.constructed` attaches listeners *after* `props.constructed`'s
+		// synchronous drain — without this re-fire, late-bound `on*` handlers
+		// miss the initial dispatch. Duplication for pre-connect listeners is
+		// the tradeoff.
```
LeaVerou: I don't understand this.

src/plugins/props/util/Prop.js
```diff
@@ -96,11 +96,7 @@ let Self = class Prop {
 		return types.parse(value, this.type);
 	}
 
-	/**
-	 * Subscriber for Computed signals (spec.get, spec.convert, spec.default).
-	 * Updates element.props cache, reflects to attributes if opted in,
-	 * and fires propchange events.
-	 */
+	// Subscriber for Computed signals (spec.get, spec.convert, default fallthrough).
 	#onComputedChange (element, source, newValue, oldValue) {
```
LeaVerou, line 103: I don't understand what this does or when it's called. Is it called when a Computed dependent changes?

src/plugins/props/util/Props.js
```diff
@@ -50,35 +55,95 @@ export default class Props extends Map {
...
+		if (!map) {
+			map = new Map();
+			this.#eventDispatchQueue.set(element, map);
 		}
+
+		let existing = map.get(prop.name);
+		if (existing) {
+			// Coalesce: latest parsedValue/source wins, but old values stay pinned
+			// to the first write so the payload spans the full first→last delta.
+			let { oldInternalValue, oldAttributeValue } = existing.detail;
```
LeaVerou: As I suggesteed in the other PR, I think `oldInternalValue` should just be `oldValue`, unless we use `value` for something else (?)

LeaVerou:
Btw I think Claude didn't reply to this at all:

> The way I was thinking about it, we'd fire propchange as usual (sync, per property), but also have a debounced `propschange` (or `propsupdate`) that fires a certain delay after the _last_ prop change (i.e. debounced).
> 
> This could live in the Props controller for a given element (perhaps make an `ElementProps` class instead of having `method(element)` on `Props`). It would listen to propchange, collect events, drain XXms after the last propchange OR wait until the element gets connected. For disconnected elements, I'd expect the low-level propchange to still fire, but not the higher level propschange. But that's not a strong opinion.
> 
> It seems that should be a fairly small and straightforward change, not ...whatever this is that Claude did. But perhaps I'm missing some factors that necessitated this complexity?

It just justified its current solution (and I agree wrt the extent of changes). But I still want to know whether there's an error in my reasoning here.

src/plugins/props/util/Props.js
DmitrySharabin: Done.

There was a name collision: `Prop.set` already destructured `oldValue` as a param—the DOM-aligned old attribute string from `Props.attributeChanged`, which mirrors `attributeChangedCallback(name, oldValue, newValue)`. The old parsed value (formerly `oldInternalValue`) wanted that same name on the change object.

Renamed the destructured param to `oldAttributeValue` instead. The DOM-aligned `oldValue` stays at the host-element boundary (`Props.attributeChanged`'s param + the `attributeChangedCallback` shim in `props/index.js`); inside `Prop.set` and on the change object, it's now `oldAttributeValue` (literally the old attribute string) and `oldValue` (the parsed value).

The same logic applied to `parsedValue` — the disambiguation prefix existed only because of a `value` field on the plain-Signal change object.

src/plugins/props/util/Prop.js
DmitrySharabin: Yes. Updated the comment. Does it read better now? Is it clear from the comment what it does now?

DmitrySharabin:
> But I still want to know whether there's an error in my reasoning here.

You're right. In terms of #51, your debounce-or-microtask + bulk-only model is sufficient (and also covers #100). This PR went beyond #51's ask — I added per-prop coalescing of `propchange` itself, motivated by `propChangedCallback({ name })` patterns in `color-elements` that read other props expecting them to be settled. That extra scope wasn't asked for.

If we move to your shape, here's what we lose at the per-prop layer:

- **Per-prop coalescing**: `el.x = 1; el.x = 2` → two `propchange` events for `x`, not one.
- **Round-trip short-circuit**: `el.x = 1; el.x = 0` fires two events even when net change is zero.
- **First→last delta semantic** on `propchange` detail (`oldValue` from first write, `value` from last) — drops to per-write semantics.

What stays:

- `propschange` (bulk) coalescing — sync writes across props collapse to one event with all changes.
- Connect-time flushing — for the bulk event only; per-prop events fire sync at write time (lost on disconnected elements, matching main).

Implementation collapses by roughly half: drop the per-element drain queue, the drain-time equality short-circuit, and reconnect-drain for per-prop events. Color-elements migrates cross-prop reads from `propChangedCallback` to `updated()` (mostly done in https://github.com/color-js/elements/pull/227).

**Two paths forward:**

1. **Full rewrite of this PR** to your shape; per-prop coalescing drops entirely.
2. **Split** — rewrite this PR to your shape (closes #51 + #100, bulk-only); move per-prop coalescing into a stacked PR on top, evaluated on its own merits. If the scope creep doesn't earn its keep, that PR closes. Mirrors the #106 / #102 / #107 stack we already have.

Near-term state is the same either way — your simpler shape lands. Which would you prefer?

src/plugins/events/propchange.js
DmitrySharabin: The comment was opaque — fair. Updated in to lead with a “don't remove” warning and describe the timing concretely:

```js
// Don't remove: re-fires initial propchange for on*= attribute handlers,
// which onprops attaches *after* the initial dispatch. Without this,
// shortcut handlers declared in HTML never see the initial event.
// Pre-connect imperative listeners receive the event twice.
```

Background: this re-fire was actually deleted in an earlier session, which broke `<x-foo onfoochange="…">` handlers in `color-elements`. The previous comment was meant to deter future deletion but didn't communicate that clearly. Does this read better?

src/plugins/props/util/Prop.js
```diff
@@ -289,33 +287,33 @@ let Self = class Prop {
...
-					let attributeName = this.toAttribute;
-					let attributeValue = this.stringify(parsedValue);
-					let oldAttributeValue = element.getAttribute(attributeName);
-
-					if (oldAttributeValue !== attributeValue) {
-						element.ignoredAttributes.add(this.toAttribute);
+			if (source === "property" && this.toAttribute) {
+				let attributeName = this.toAttribute;
+				let attributeValue = this.stringify(parsedValue);
+				let oldAttributeValue = element.getAttribute(attributeName);
```
LeaVerou: A lot of this diff is cosmetic changes and renames which makes it hard to review the actual changes…

src/plugins/props/util/PropsChangeEvent.js
```diff
@@ -0,0 +1,7 @@
+export default class PropsChangeEvent extends CustomEvent {
+	constructor (changedProps, options) {
```
LeaVerou: I would make `changedProps` an option, we shouldn't be changing the constructor shape on subclasses without good reason.

LeaVerou:
> > But I still want to know whether there's an error in my reasoning here.
> 
> You're right. In terms of #51, your debounce-or-microtask + bulk-only model is sufficient (and also covers #100). This PR went beyond #51's ask — I added per-prop coalescing of `propchange` itself, motivated by `propChangedCallback({ name })` patterns in `color-elements` that read other props expecting them to be settled. That extra scope wasn't asked for.
> 
> If we move to your shape, here's what we lose at the per-prop layer:
> 
> * **Per-prop coalescing**: `el.x = 1; el.x = 2` → two `propchange` events for `x`, not one.
> * **Round-trip short-circuit**: `el.x = 1; el.x = 0` fires two events even when net change is zero.
> * **First→last delta semantic** on `propchange` detail (`oldValue` from first write, `value` from last) — drops to per-write semantics.
> 
> What stays:
> 
> * `propschange` (bulk) coalescing — sync writes across props collapse to one event with all changes.
> * Connect-time flushing — for the bulk event only; per-prop events fire sync at write time (lost on disconnected elements, matching main).
> 
> Implementation collapses by roughly half: drop the per-element drain queue, the drain-time equality short-circuit, and reconnect-drain for per-prop events. Color-elements migrates cross-prop reads from `propChangedCallback` to `updated()` (mostly done in [color-js/elements#227](https://github.com/color-js/elements/pull/227)).
> 
> **Two paths forward:**
> 
> 1. **Full rewrite of this PR** to your shape; per-prop coalescing drops entirely.
> 2. **Split** — rewrite this PR to your shape (closes [[Props] Bunch up property changes #51](https://github.com/nudeui/element/issues/51) + [`propchange` events queued while disconnected are not dispatched on reconnect #100](https://github.com/nudeui/element/issues/100), bulk-only); move per-prop coalescing into a stacked PR on top, evaluated on its own merits. If the scope creep doesn't earn its keep, that PR closes. Mirrors the [Populate detail on first_connected shortcut re-fires #106](https://github.com/nudeui/element/pull/106) / Batch propchange events on a microtask drain #102 / [Defaults don't reflect to attributes (closes #105) #107](https://github.com/nudeui/element/pull/107) stack we already have.
> 
> Near-term state is the same either way — your simpler shape lands. Which would you prefer?

We definitely want coalescing! That's not scope creep, that's part of implementing #51 correctly.

Though it seems to me that coalescing would happen organically _anyway_ given that we pass a map to `updated()`, the only thing we'd need for coalescing is to run all new values through `equals()` before passing the map to any consumers, which we should be doing _anyway_ (and checking if `size` is > 0 at the end. of this). But yes, we'd need to track initial values at the start of the changes and compare against those.

The current diff LGTM too, I just need to make. surewe're not overengineering this.

src/plugins/props/util/Prop.js
DmitrySharabin: Let me restore the previous code structure to make it easier.

DmitrySharabin: Done

src/plugins/props/util/PropsChangeEvent.js
DmitrySharabin: Fixed. I also made it aligned with the `PropChangeEvent` class and the `CustomEvent` signature, in general.

src/plugins/props/util/Prop.js
```diff
@@ -305,17 +301,22 @@ let Self = class Prop {
 						element.ignoredAttributes.add(this.toAttribute);
 
 						Object.assign(change, { attributeName, attributeValue, oldAttributeValue });
-						this.applyChange(element, { ...change, source: "attribute" });
+						if (attributeValue === null) {
+							element.removeAttribute(attributeName);
+						}
+						else {
+							element.setAttribute(attributeName, attributeValue);
```
LeaVerou: Use `toggleAttribute()`

src/plugins/props/util/Props.js
```diff
@@ -1,7 +1,12 @@
 import Prop from "./Prop.js";
 import PropChangeEvent from "./PropChangeEvent.js";
+import PropsChangeEvent from "./PropsChangeEvent.js";
 
 export default class Props extends Map {
+	#eventDispatchQueue = new WeakMap();
+	#pendingElements = new Set();
+	#drainScheduled = false;
```
LeaVerou, line 27: These need jsdoc to explain what they're for

DmitrySharabin: Done

src/plugins/props/util/Prop.js
DmitrySharabin: Honestly, I don't understand how I can use `toggleAttribute` here. It doesn't support passing the value as an argument. 🤷‍♂️

LeaVerou (approved):

PR #107, opened 2026-05-01
Convergent restore-default for reflected props

DmitrySharabin:
## Summary

Builds on #108 (regression fixes) to add the **clear-stale-attribute** behavior for reflected props. After this PR, `el.prop = undefined` and `el.removeAttribute(name)` converge — both revert the prop to the default *and* clear any previously-reflected attribute.

> Stacked on [#102](https://github.com/nudeui/element/pull/102). #108 lands first.

## Behavior

| Action | After #108 | After this PR |
|---|---|---|
| `el.bar = X` (where X ≠ default) | el.bar=X, attr=`"X"` | el.bar=X, attr=`"X"` |
| `el.bar = X` where X = default | el.bar=X, attr=`"X"` | el.bar=X, attr=`"X"` |
| `el.bar = undefined` after a prior write | el.bar=default, **attr stays X** | el.bar=default, **attr=null** |
| `el.removeAttribute("bar")` after a prior write | el.bar=default, attr=null | el.bar=default, attr=null |
| Reactive default re-evaluates while a stale attribute lingers | attr stays X | attr=null |

The "stays X" rows are a latent bug present even on pre-signals (the old `applyChange` helper's `?? element.getAttribute(...)` treats explicit `null` as "use fallback" and silently leaks a stale attribute). Not technically a regression — but worth fixing alongside the convergent-restore contracts.

## Implementation

- **`#reflect(element, value, source)`** — extends the helper #108 introduces. Maps `source === "default"` to a `null` attribute target; the existing "write if different" logic does the rest, naturally clearing stale attributes without a separate code path.
- **`#settle(element, value)`** — single source of truth for the default → parse → convert pipeline. Used from the Computed body and from `Prop#set`'s rawSignal path so the inline reflection (introduced in #108) sees the same settled value.

## Trade-offs

- **Reflection is decoupled from the `propchange` event.** Explicit writes always reflect; events follow value-change semantics. Writing the current default-resolved value reflects to the DOM but doesn't fire `propchange` (consistent with the renamed `Assigning current default-resolved value is a no-op for events` test).
- **`el.prop = null` is preserved as a legitimate user value** (mirrors JS conventions). Use `undefined` or `removeAttribute()` to restore the default.

## Test plan

- [x] `npm test -- --ci` — 101/102 PASS, 0 FAIL, 1 pre-existing skip in `split()`
- [x] All 4 scenarios from the [review comment](https://github.com/nudeui/element/pull/107#issuecomment-4371809054) exercised in tests

## Rebase note

This branch's current diff still includes parts of #108 (the regression fixes). Once #108 lands and this branch rebases on the new main, those commits dedupe automatically. The substantive remainder is `1c8149c` (data-flow encoding for stale-attribute clearing), `568c6e2` (`#settle` extraction), `e3f2180` (docs), and the test additions.

### DIFF
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -165,6 +165,7 @@ let Self = class Prop {
 				let rawSignal = new Signal(undefined, options);
 				this.#rawSignals.set(element, rawSignal);
 
+				let source = this.spec.convert ? "convert" : "property";
 				signal = new Computed(() => {
 					let value = rawSignal.value;
 					if (value === undefined && this.spec.default !== undefined) {
@@ -188,14 +189,8 @@ let Self = class Prop {
 					return value;
 				}, options);
 				signal.subscribe((newValue, oldValue) => {
-					let source =
-						rawSignal.value === undefined
-							? "default"
-							: this.spec.convert
-								? "convert"
-								: "property";
-
-					this.#onComputedChange(element, source, newValue, oldValue);
+					let newSource = rawSignal.value === undefined ? "default" : source;
+					this.#onComputedChange(element, newSource, newValue, oldValue);
 				});
 			}
 			else {
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -105,10 +105,10 @@ let Self = class Prop {
 	#onComputedChange (element, source, newValue, oldValue) {
 		element.props[this.name] = newValue;
 
-		// Skip reflection for default values: synthesized attributes would
-		// clobber programmatic writes that landed before initializeFor (#105).
-		if (this.toAttribute && source !== "default") {
-			let attributeValue = this.stringify(newValue);
+		// `null` for defaults: don't synthesize an attribute (would clobber pre-mount
+		// writes — #105) and clear any stale attribute left over from a prior write.
+		if (this.toAttribute) {
+			let attributeValue = source === "default" ? null : this.stringify(newValue);
 			let oldAttributeValue = element.getAttribute(this.toAttribute);
 			if (oldAttributeValue !== attributeValue) {
 				element.ignoredAttributes.add(this.toAttribute);
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -278,6 +278,13 @@ let Self = class Prop {
 			return;
 		}
 
+		// removeAttribute() arrives as null; collapse to undefined so the prop
+		// reverts to its natural empty state (default, if any, otherwise just
+		// undefined). Property writes of null remain a legitimate user value.
+		if (source === "attribute" && parsedValue === null) {
+			parsedValue = undefined;
+		}
+
 		if (this.equals(parsedValue, oldInternalValue)) {
 			return;
 		}
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -146,7 +146,8 @@ let Self = class Prop {
 		let signal = this.#signals.get(element);
 
 		if (!signal) {
-			// `forceNotify` so explicit user writes still reach the subscriber when
+			// Delegate equality to the prop's type-aware equality.
+			// Explicit user writes still reach the subscriber when
 			// the Computed dedupes against a cached default-resolved value (#105).
 			let options = { equals: (a, b) => this.equals(a, b), forceNotify: true };
 
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -121,7 +121,7 @@ let Self = class Prop {
 			}
 		}
 
-		// Gate event on value change (the Computed uses forceNotify).
+		// Gate event on value change (the Computed uses force).
 		if (this.equals(newValue, oldValue)) {
 			return;
 		}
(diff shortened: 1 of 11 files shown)


src/plugins/props/util/Prop.js
```diff
@@ -97,11 +97,12 @@ let Self = class Prop {
 	}
 
 	// Subscriber for Computed signals (spec.get, spec.convert, default fallthrough).
-	#onComputedChange (element, source, newValue, oldValue) {
+	#onComputedChange (element, source, newValue, oldValue, fromDefault = false) {
```
LeaVerou: Wait, why do we need a `fromDefault`? Isn't this what `source: "default"` is for?

src/plugins/props/util/Prop.js
```diff
@@ -177,7 +178,8 @@ let Self = class Prop {
 					return value;
 				}, options);
 				signal.subscribe((newValue, oldValue) => {
-					this.#onComputedChange(element, source, newValue, oldValue);
+					let fromDefault = rawSignal.value === undefined;
```
LeaVerou: This reminds me, **defaults** are one case where `updated()` consumers may want to read *either* the internal value *or* the getter value. They can always get the current getter value via the actual getter, but not the old value.

src/plugins/props/util/Prop.js
DmitrySharabin: You're right. Refactored to make `source` dynamic instead of adding a flag.

> [!NOTE]
> For default/convert-backed props, `source` reports the **pipeline that produced the value** (`"default"` / `"convert"` / `"property"`). Internally consistent if you read `source` as “what produced this value” rather than “how did the user trigger it.” So `el.setAttribute("foo", "X")` on a default-backed prop still surfaces as `"property"`, not `"attribute"`—there's a Computed wrapper running, and the pipeline is what produced the value. `"attribute"` only applies to plain Signal props where there's no pipeline.

LeaVerou:
Before I review the code, one thing I want to make sure (and we should have tests for) is that the behavior is as expected.

Suppose you have element `<x-foo>` with attribute `bar` with a default value of `5` that reflects both ways.

- `<x-foo></x-foo>` → should not reflect, `foo.bar` returns `5`
- `<x-foo id=foo></x-foo>` with JS doing `foo.bar = 5;` *should* reflect (you see `<x-foo bar="5"></x-foo>` in the DOM)
- `foo.bar = undefined; foo.bar` should return `undefined`, not `5` (though if that complicates implementation, we can revisit)
- `<x-foo id=foo bar="6"></x-foo>` and then `foo.removeAttribute("bar"); foo.bar;` → returns `5`

DmitrySharabin:
I added the corresponding tests. I just would like to discuss one case.

Let's consider the following definition of `<x-foo>`:

```js
import NudeElement from "nude-element";

class XFoo extends NudeElement {
	static props = {
		bar: {
			type: Number,
			default: 5,
			reflect: true, // spelled out for clarity
		},
	};
}

customElements.define("x-foo", XFoo);
```

> * `<x-foo></x-foo>` → should not reflect, `foo.bar` returns `5`

This is how it works now. No changes required.

> * `<x-foo id=foo></x-foo>` with JS doing `foo.bar = 5;` _should_ reflect (you see `<x-foo bar="5"></x-foo>` in the DOM)

Because `bar` has a default value, its value is stored in `rawSignal` and wrapped by a computed signal. After the element is mounted, the computed signal has emitted 5 (from default fallthrough). Then `foo.bar = 5` writes `5` into `rawSignal`, and the computed one re-evaluates to `5`. But the computed's `equals(5, 5)` is `true`, so it short-circuits and never notifies its subscriber. That subscriber is what drives both reflection and the `propchange` event, so neither happens.

With the recent changes, it also works as you described. The reflection is decoupled from the computed dedupe via an inline path in `Prop#set`, so the user's write reaches the DOM. The `propchange` event, however, still follows value-change semantics: writing the current value is a no-op for events.

> * `foo.bar = undefined; foo.bar` should return `undefined`, not `5` (though if that complicates implementation, we can revisit)

Writing `foo.bar = undefined` clears `rawSignal` back to `undefined`, which the computed wrapper then resolves to `5` via the default. This is also what makes `defaultProp` work since it desugars to `default() { return this[other] }` and relies on writing `undefined` to restore the link after an explicit write severs it (see the [“defaultProp restores on undefined”](https://github.com/nudeui/element/blob/1984e21d31b82f1c463926cf35ceec766666823c/test/Prop.js#L398-L412) test).

Returning `undefined` in this case would require a separate “user explicitly set to undefined” sentinel—distinct from “rawSignal is at its initial state”. This is a bigger change that would also need a new way for `defaultProp` consumers to re-engage the link.

> * `<x-foo id=foo bar="6"></x-foo>` and then `foo.removeAttribute("bar"); foo.bar;` → returns `5`

It also works with the recent changes.

LeaVerou:
Yeah, the important bit is to distinguish "explicitly set to default value" from "set to default value because no value has been set". E.g. a button component may default to `variant=neutral` (and in the CSS you may treat it as "neutral unless something else is inherited"), but that's fundamentally different from explicitly setting its `variant` to `neutral`.

`undefined` is less important, since components can always use another value to mean empty, so it's ok if `undefined` is always converted to the default value — I think. If anything, it gives consumers an explicit way to reset to the default value.

DmitrySharabin:
Superseded by #108 — the stale-attribute clearing was bundled into the same default-reflection logic in #108. The simplified diff is a 4-line tweak in #onComputedChange (#108) plus a regression test.

PR #108, opened 2026-05-06
Restore pre-signals defaults / reflection behavior (closes #105)

### LINKED ISSUE #105: Race: external read of a Computed-backed prop before `initializeFor` pollutes the attribute and clobbers programmatic writes
## Summary

A `Computed`-backed prop's lazy first compute fires its subscriber, which reflects the resolved value to the attribute. If this first compute is triggered by an **external read** (e.g. a parent component reading a child's prop during the parent's own mount) **before** the child's `Props.initializeFor` runs, then:

1. The resolved default is reflected to the child's attribute.
2. A programmatic write to the child between that read and the child's `initializeFor` writes to `rawSignal` but the synthesized attribute remains.
3. The child's `Props.initializeFor` Pass 1 reads the attribute (now holding the synthesized default), treats it as user input, and overwrites the programmatic write.

Net result: the programmatic write is silently lost.

## Repro (drop-in test)

Add this to the "Runtime behavior" group in [`test/Prop.js`](test/Prop.js), alongside the "Attribute reflection" sub-group:

```js
{
    name: "Initialization timing",
    tests: [
        {
            name: "Programmatic write between an external read and initializeFor wins over default reflection",
            async run () {
                let Class = FakeElement.with({
                    v: { type: Number, default: 7, reflect: true },
                });
                let el = new Class();

                void el.v;   // External read: forces lazy first compute
                el.v = 42;   // Programmatic write before mount

                el.mount();  // Child's initializeFor runs after the parent's logic
                await flush(2);

                return [el.v, el.getAttribute("v")];
            },
            expect: [42, "42"],
        },
    ],
},
```

Currently returns `[7, "7"]` on `main` and on `props-batched-drain` — the synthesized default attribute is re-imported by Pass 1, overwriting `rawSignal := 42` back to `7`.

## Real-world surface

Found while integration-testing `nude-element` against [`color-elements`](https://github.com/leaverou/color-elements). The `<color-picker>` element contains a `<space-picker>` in its shadow DOM. During color-picker's own mount its `spaceId.changed` propchange handler reads `this._el.space_picker.value` ([color-picker.js:149-152](https://github.com/leaverou/color-elements/blob/main/src/color-picker/color-picker.js#L149-L152)), forcing space-picker's `Computed` to first-compute and reflect the default `"a98rgb"`. Color-picker then pushes `sp.value = "lab"`. Space-picker's `Props.initializeFor` runs next, sees the synthesized `value="a98rgb"` attribute, and overwrites `rawSignal` back to `"a98rgb"` — clobbering the parent's intent.

## Detailed trace

1. Color-picker reads `sp.value` → lazy `getSignal` creates a `Computed` with its subscriber attached at construction time. `signal.value` triggers first compute → falls through to default `"a98rgb"`.
2. Subscriber `#onComputedChange` fires synchronously: `setAttribute("value", "a98rgb")` on the inner space-picker (wrapped in `ignoredAttributes.add/delete` — but that only protects the synchronous `attributeChangedCallback` window).
3. Color-picker writes `sp.value = "lab"` → `prop.set` writes `rawSignal := "lab"`. The Computed marks itself dirty; its subscriber hasn't fired yet (microtask flush).
4. Space-picker's `Props.initializeFor` runs (next microtask). Pass 1 iterates `observedAttributes` and finds `value="a98rgb"` (synthesized in step 2) → `prop.set(el, "a98rgb", { source: "attribute" })` → `rawSignal := "a98rgb"`. **Programmatic write lost.**

The `ignoredAttributes` mechanism doesn't catch this because Pass 1 reads attributes via `getAttribute()` directly (not via `attributeChangedCallback`), and by then the ignored set is empty.

## Why this is hard to fix cleanly

There's a contract conflict to resolve before picking an implementation. [`test/Prop.js`](test/Prop.js) codifies the following behaviors:

- **"Default + reflect reflects on mount (plain)"** — `{ default: 7, reflect: true }` → `el.getAttribute("plain") === "7"` after mount.
- **"Default + convert + reflect reflects on mount"** — same, with convert applied.
- **"Reflection writes go to reflect.to alias, not the prop name"** — expects `[["data-v", "7"], ["data-v", "42"]]` (default reflects on mount, then update).

These tests assert that **defaults DO materialize as attributes** on mount. That's the explicit, opt-in contract. The race surfaces because the same machinery materializes defaults at the *wrong* time as well.

### Option A — change the contract ("defaults don't reflect")

- **Implementation**: gate reflection in `#onComputedChange` on whether `rawSignal` has a user-set value:
  ```js
  let fromDefault = rawSignal.value === undefined;
  this.#onComputedChange(element, source, newValue, oldValue, fromDefault);
  ```
- **Pros**: smallest diff (~10 lines).
- **Cons**:
  - Breaks the 3 tests above and the contract they encode.
  - Leaves stale attributes when a user writes `undefined` to re-engage the default after a prior write (subscriber fires with `rawSignal.value === undefined` → skips reflection → previously-set attribute remains). The "matches native HTML" justification doesn't fully hold even within the new code.
  - Requires a README note documenting the breaking behavior change.

### Option B1 — per-element initialization flag (minimal, contract-preserving)

- `Props` tracks `#initialized = new WeakSet()`. `initializeFor(element)` adds the element at the **start** (before Pass 1).
- `#onComputedChange` gates the reflection block on `this.props.hasInitialized(element)`.
- **Pros**: ~6 lines net. Preserves all 3 contract tests. Correctly handles re-engaging default after a write (no stale attribute).
- **Cons**: Still has a "propchange blip" — early read fires `propchange("a98rgb")`, later write fires `propchange("lab")`. Two events for what's logically one transition. Not a regression vs. Option A but architecturally ugly.

### Option B2 — subscribe-after-init (architectural, preferred)

The deepest root cause is that the `Computed` subscriber owns work (`element.props` update, reflection, `propchange`) that should belong to `initializeFor`. A first compute is initialization, not a value change.

- **Implementation**:
  - Remove `signal.subscribe(...)` from both Computed branches of `getSignal`.
  - In `prop.initializeFor`, after the existing logic: force-read `signal.value`, manually call `#onComputedChange` once (if `element.props[name]` is undefined and value is defined), then subscribe for subsequent changes.
  - Move the closured `source` string to a per-Prop instance field so `prop.initializeFor` can pass it to `#onComputedChange`.
  - Add `#subscribed = new WeakSet()` and `#computedSource` private fields on `Prop`.
- **Pros**:
  - Eliminates the race entirely — subscribers physically can't fire before init.
  - Eliminates the propchange blip (only one `propchange` fires in the bug scenario, with the settled value).
  - Preserves the default link in all cases (including write-undefined-to-re-engage).
  - Clean separation: `initializeFor` does init, the subscriber does reactivity.
- **Cons**: ~15-line refactor; touches the lazy-subscribe pattern. Worth its own review pass.

### Option C — workaround in color-elements

- Restructure `color-picker`'s `spaceId.changed` to not read `sp.value` synchronously during the parent's drainFor. E.g. defer to a microtask, or compare against a last-known-pushed value tracked in `color-picker` itself.
- **Pros**: zero `nude-element` changes.
- **Cons**: papers over a real `nude-element` bug; other consumers can hit the same race in different shapes; doesn't surface the architectural issue.

## Concerns to track regardless of which fix lands

- **`propchange` "blip"**: two events for one logical transition. Only Option B2 eliminates it.
- **Default-link severance**: should an external read before `initializeFor` sever the default link?
  - B2: no (rawSignal stays undefined throughout the early read).
  - B1: same as B2 (subscriber fires but doesn't reflect; rawSignal still undefined).
  - A: depends on timing of subsequent attribute import.
- **`Object.hasOwn` instance-value path** ([Prop.js:201-209](src/plugins/props/util/Prop.js#L201-L209)): pre-existing quirk where the resulting propchange `source` reports the closured `"default"`/`"convert"` rather than `"property"`. All three options inherit this; not a regression.
- **`spec.get` Computeds with `reflect:true`**: `this.reflect = spec.reflect ?? !this.spec.get` — reflection defaults off for `get`-props. Any fix should preserve that default while allowing opt-in.
- **Re-mount after disconnect**: WeakSet entries (B1, B2) persist across disconnect/reconnect. That's almost certainly correct (the prop layer doesn't "uninitialize"), but worth noting.

## Recommended next steps

1. Decide between A / B1 / B2 (preference: **B2**).
2. Land the failing test (drop-in code above) on the working branch.
3. Implement the chosen fix.
4. Add a prose note in [`src/plugins/props/README.md`](src/plugins/props/README.md) clarifying when defaults reflect (whatever the final rule is).
5. Consider documenting the `Object.hasOwn` source-reporting quirk separately.

## Background

A drive-by patch implementing Option A was tested while debugging the `color-elements` integration. It fixed the symptom but broke the 3 contract tests above without test/doc updates, so it was rejected.

The race is a **pre-existing bug** from #91 (signals-props) — not introduced by #102 (props-batched-drain). Filed separately so #102 can ship focused on its actual scope.

DmitrySharabin:
Closes #105.

Restore three default-reflection contracts that the signals migration (#91) regressed. While editing the same `#onComputedChange` conditional, also clear a long-standing latent bug — leaving it would have left #105 partially open. The production color-elements site runs on 0.1.3 (pre-signals); these restore the contracts it relies on, so #102 can ship without regressions in its diff.

## What's restored

| Scenario | Pre-signals | Post-signals (origin/main) | This PR |
|---|---|---|---|
| Default reflects on mount | no | **yes (regression)** | no |
| `removeAttribute(name)` → `el.prop` | default | **null (regression)** | default |
| Explicit `el.prop = X` where `X` matches the default | reflects `"X"` | **no reflection (regression)** | reflects `"X"` |
| Reverting to default leaves a stale reflected attribute from a prior write | **yes (latent)** | yes | no |

The 4th row is a latent bug present even pre-signals: `applyChange`'s `?? element.getAttribute(...)` made explicit `null` mean "fall back to current attribute," so callers had no way to express "remove this attribute" through the change object. This PR fixes both layers — `applyChange` is tightened so callers can actually request removal, and the `rawSignal` subscriber passes `null` when reverting to default.

## Code shape: pre-signals' reflection scope

Pre-signals' `set()` only reflected when `source === "property"` — direct prop writes. Convert-dep, get-dep, and default-dep cascades went through `update()` → `set()` with non-`"property"` sources and fell through the reflection gate. The signals migration moved reflection onto `#onComputedChange` (the Computed's subscriber), which fires on every recompute regardless of trigger. That expansion is the root of #105 (default fallthrough started synthesizing attributes) and of the user-write-equal-to-default reflection gap (the Computed's `equals` dedup skips recomputes whose new value equals the cached value).

This PR moves reflection to a subscriber on `rawSignal`. `rawSignal.value` changes for every `prop.set` call on a Computed-backed prop — user writes (`source: "property"`) and attribute-driven writes (`source: "attribute"`) — and *only* those. That's exactly pre-signals' scope. `#onComputedChange` no longer touches attributes; it just updates `element.props` and fires `propchange`.

## Stats

GitHub shows **+104 / −43 across 5 files**. Most of the diff is tests:

| Category | Net lines | File(s) |
|---|---:|---|
| Source code logic + JSDoc | +10 | Prop.js (+21), signals.js (−11) |
| Documentation | +2 | README.md |
| Adjusted + new regression tests | +45 | test/Prop.js |
| Test infrastructure | +4 | test/util/FakeElement.js (`setAttribute` / `removeAttribute` now call `attributeChanged`, mirroring real custom elements) |

`src/signals.js` is **net −11 lines** — no new Signal API. The redundant equals guard in `Computed#compute` is removed (`Signal#set` decides), and the previous prototype-descriptor walk is replaced with `super.value = value`.

Each new test was confirmed to **fail with its corresponding fix disabled**, then **pass with the fix re-enabled**.

## Implementation

- **`rawSignal.subscribe(...)` in `getSignal`**: the single reflection site. Fires on direct prop writes and attribute-driven writes — the same trigger set pre-signals reflected on. When the raw value is `null`/`undefined` (revert to default or explicit `null` write), clears the attribute; otherwise applies `spec.convert` and reflects the stringified result. The `getAttribute === attributeValue` guard short-circuits attribute-driven writes (the value is already there) and any other no-op.
- **`#onComputedChange`**: only updates `element.props[name]` and fires `propchange`. No reflection.
- **Source derivation**: Compute `source` dynamically in the Computed's subscriber — `"default"` iff `rawSignal.value === undefined`, else the user-write source (`"convert"` or `"property"`).
- **`Prop#set`**: Collapse `null → undefined` when `source === "attribute"`, so `removeAttribute` returns the rawSignal to its empty state and the Computed re-engages the default fallthrough. The `rawSignal` branch is now a single line — the previous inline `#reflect` workaround is gone (the subscriber handles it).
- **`applyChange` `??` → `!== undefined ?`**: Closes the latent bug above. Now explicit `null` means "remove"; only an omitted `attributeValue` falls back to the source element's current attribute.
- **`Computed.#compute`**: writes via `super.value = value` (standard class inheritance) instead of `Object.getOwnPropertyDescriptor(Signal.prototype, "value").set.call(this, value)`. Same observable behavior, no internals interference.

## Why now

PR #102 currently inherits the regressions. Landing this first lets #102 rebase cleanly on a fixed main, so it ships without regressions in its diff.

## Test plan

- [x] `npm test -- --ci` — 85/88 PASS, 3 pre-existing skips, 0 FAIL

### DIFF
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -104,9 +104,10 @@ let Self = class Prop {
 	#onComputedChange (element, source, newValue, oldValue) {
 		element.props[this.name] = newValue;
 
-		// `null` for defaults: don't synthesize an attribute (would clobber pre-mount
-		// writes — #105) and clear any stale attribute left over from a prior write.
+		// Reflect to attribute if this prop opts in
 		if (this.toAttribute) {
+			// Don't synthesize attributes for defaults (it would clobber pre-mount writes; see #105).
+			// Clear any stale attribute left over from a prior write.
 			let attributeValue = source === "default" ? null : this.stringify(newValue);
 			let oldAttributeValue = element.getAttribute(this.toAttribute);
 			if (oldAttributeValue !== attributeValue) {
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -165,6 +165,7 @@ let Self = class Prop {
 				let rawSignal = new Signal(undefined, options);
 				this.#rawSignals.set(element, rawSignal);
 
+				let source = this.spec.convert ? "convert" : "property";
 				signal = new Computed(() => {
 					let value = rawSignal.value;
 					if (value === undefined && this.spec.default !== undefined) {
@@ -188,14 +189,8 @@ let Self = class Prop {
 					return value;
 				}, options);
 				signal.subscribe((newValue, oldValue) => {
-					let source =
-						rawSignal.value === undefined
-							? "default"
-							: this.spec.convert
-								? "convert"
-								: "property";
-
-					this.#onComputedChange(element, source, newValue, oldValue);
+					let newSource = rawSignal.value === undefined ? "default" : source;
+					this.#onComputedChange(element, newSource, newValue, oldValue);
 				});
 			}
 			else {
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -278,6 +278,13 @@ let Self = class Prop {
 			return;
 		}
 
+		// removeAttribute() arrives as null; collapse to undefined so the prop
+		// reverts to its natural empty state (default, if any, otherwise just
+		// undefined). Property writes of null remain a legitimate user value.
+		if (source === "attribute" && parsedValue === null) {
+			parsedValue = undefined;
+		}
+
 		if (this.equals(parsedValue, oldInternalValue)) {
 			return;
 		}
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -146,7 +146,8 @@ let Self = class Prop {
 		let signal = this.#signals.get(element);
 
 		if (!signal) {
-			// `forceNotify` so explicit user writes still reach the subscriber when
+			// Delegate equality to the prop's type-aware equality.
+			// Explicit user writes still reach the subscriber when
 			// the Computed dedupes against a cached default-resolved value (#105).
 			let options = { equals: (a, b) => this.equals(a, b), forceNotify: true };
 
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -121,7 +121,7 @@ let Self = class Prop {
 			}
 		}
 
-		// Gate event on value change (the Computed uses forceNotify).
+		// Gate event on value change (the Computed uses force).
 		if (this.equals(newValue, oldValue)) {
 			return;
 		}
@@ -151,7 +151,7 @@ let Self = class Prop {
 			// Delegate equality to the prop's type-aware equality.
 			// Explicit user writes still reach the subscriber when
 			// the Computed dedupes against a cached default-resolved value (#105).
-			let options = { equals: (a, b) => this.equals(a, b), forceNotify: true };
+			let options = { equals: (a, b) => this.equals(a, b), force: true };
 
 			if (this.spec.get) {
 				signal = new Computed(() => this.spec.get.call(element), options);
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -122,14 +122,16 @@ let Self = class Prop {
 		}
 
 		// Gate event on value change (the Computed uses forceNotify).
-		if (!this.equals(newValue, oldValue)) {
-			this.changed(element, {
-				element,
-				source,
-				parsedValue: newValue,
-				oldInternalValue: oldValue,
-			});
+		if (this.equals(newValue, oldValue)) {
+			return;
 		}
+
+		this.changed(element, {
+			element,
+			source,
+			parsedValue: newValue,
+			oldInternalValue: oldValue,
+		});
 	}
 
 	/**
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -122,7 +122,7 @@ let Self = class Prop {
 			}
 		}
 
-		// Gate event on value change (the Computed uses force).
+		// Gate event on value change (the Computed uses notifyOnEquals).
 		if (this.equals(newValue, oldValue)) {
 			return;
 		}
@@ -152,7 +152,7 @@ let Self = class Prop {
 			// Delegate equality to the prop's type-aware equality.
 			// Explicit user writes still reach the subscriber when
 			// the Computed dedupes against a cached default-resolved value (#105).
-			let options = { equals: (a, b) => this.equals(a, b), force: true };
+			let options = { equals: (a, b) => this.equals(a, b), notifyOnEquals: true };
 
 			if (this.spec.get) {
 				signal = new Computed(() => this.spec.get.call(element), options);
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -104,9 +104,10 @@ let Self = class Prop {
 	#onComputedChange (element, source, newValue, oldValue) {
 		element.props[this.name] = newValue;
 
-		// Reflect to attribute if this prop opts in
+		// `null` for defaults: don't synthesize an attribute (would clobber pre-mount
+		// writes — #105) and clear any stale attribute left over from a prior write.
 		if (this.toAttribute) {
-			let attributeValue = this.stringify(newValue);
+			let attributeValue = source === "default" ? null : this.stringify(newValue);
 			let oldAttributeValue = element.getAttribute(this.toAttribute);
 			if (oldAttributeValue !== attributeValue) {
 				element.ignoredAttributes.add(this.toAttribute);
@@ -157,7 +158,6 @@ let Self = class Prop {
 				let rawSignal = new Signal(undefined, options);
 				this.#rawSignals.set(element, rawSignal);
 
-				let source = this.spec.convert ? "convert" : "default";
 				signal = new Computed(() => {
 					let value = rawSignal.value;
 					if (value === undefined && this.spec.default !== undefined) {
@@ -181,6 +181,13 @@ let Self = class Prop {
 					return value;
 				}, options);
 				signal.subscribe((newValue, oldValue) => {
+					let source =
+						rawSignal.value === undefined
+							? "default"
+							: this.spec.convert
+								? "convert"
+								: "property";
+
 					this.#onComputedChange(element, source, newValue, oldValue);
 				});
 			}
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -121,12 +121,15 @@ let Self = class Prop {
 			}
 		}
 
-		this.changed(element, {
-			element,
-			source,
-			parsedValue: newValue,
-			oldInternalValue: oldValue,
-		});
+		// Gate event on value change (the Computed uses forceNotify).
+		if (!this.equals(newValue, oldValue)) {
+			this.changed(element, {
+				element,
+				source,
+				parsedValue: newValue,
+				oldInternalValue: oldValue,
+			});
+		}
 	}
 
 	/**
@@ -143,8 +146,9 @@ let Self = class Prop {
 		let signal = this.#signals.get(element);
 
 		if (!signal) {
-			// Delegate equality to the prop's type-aware equality.
-			let options = { equals: (a, b) => this.equals(a, b) };
+			// `forceNotify` so explicit user writes still reach the subscriber when
+			// the Computed dedupes against a cached default-resolved value (#105).
+			let options = { equals: (a, b) => this.equals(a, b), forceNotify: true };
 
 			if (this.spec.get) {
 				signal = new Computed(() => this.spec.get.call(element), options);
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -146,17 +146,11 @@ let Self = class Prop {
 						return;
 					}
 
+					let resolved = this.spec.convert
+						? this.spec.convert.call(element, value)
+						: value;
 					let attributeName = this.toAttribute;
-					let attributeValue;
-					if (value == null) {
-						attributeValue = null;
-					}
-					else {
-						let resolved = this.spec.convert
-							? this.spec.convert.call(element, value)
-							: value;
-						attributeValue = this.stringify(resolved);
-					}
+					let attributeValue = this.stringify(resolved);
 
 					if (element.getAttribute(attributeName) === attributeValue) {
 						return;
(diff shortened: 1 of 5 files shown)


src/signals.js
```diff
@@ -69,6 +74,9 @@ export class Signal extends EventTarget {
 
 	set value (v) {
 		if (this.equals(v, this.#value)) {
+			if (this.#force) {
```
LeaVerou: Given that the whole point of `equals` is to determine whether to notify, wouldn't it be easier to just set an `equals` that always returns `false`? Does it _do_ anything else?

DmitrySharabin: There's actually one thing that it does additionally. It gates the cache update (e.g., on derived values). When `equals(new, current)` is `true`, the setter returns early, and `#value` stays the same (we don't reach L84). If an author uses a tolerance-based equals (for example, `(a, b) => Math.abs(a - b) < 0.1`), by replacing their function with `() => false`, we'll break the author's logic. We already have a regression test for this in [test/Prop.js#L860](https://github.com/nudeui/element/blob/fix-105-pr/test/Prop.js#L860) in the upcoming PR.

LeaVerou: Ok, then we should rename the option, because `force` as a constructor option doesn't make a lot of sense. Force what?

DmitrySharabin: Yeah, that was bothering me as well. How about `notifyOnEquals`? It's aligned with the `equals` option and explicitly expresses what it's meant to.

LeaVerou:
> Signal force option: Subscribers fire on every dep-driven recomputation, even when equals reports no change. The cache still respects equals (preserves the spec.equals contract). Reflection now tracks user intent rather than Computed dedupe; #onComputedChange gates the propchange event on actual value change.

Just realized I don't actually understand the rationale here. Why do we want to force notify even when the values are equal? It worries me we may be doing it because there is some very specific case where we need `notify` to fire, and Claude took the nuclear option instead…

src/plugins/props/util/Prop.js
```diff
@@ -280,6 +290,26 @@ let Self = class Prop {
 			// and its subscriber (#onComputedChange) handles element.props,
 			// reflection, and events.
 			rawSignal.value = parsedValue;
+
+			// Computed dedupes writes equal to the cached default (#105); reflect inline.
+			// Convert manually — `signal.value` would sync-compute and break batching.
```
LeaVerou: I don't understand what this block of code does, and the comment isn't helping.

DmitrySharabin: > [from Claude — replying on @dmitrysharabin's behalf]

You're right that the comment didn't justify why this block exists. Two changes since your review:

1. **Extracted `#reflect(element, value)` as the single reflection site**, called from both `#onComputedChange` (dep-driven path) and this user-write path. This restores the shape pre-signals had: one place where attribute reflection happens, called from both the user-write path (`set()`) and the dep-driven path (`update()` going through `set()`). The signals migration accidentally split it in two — collapsing back to a shared helper makes this site a one-line delegation, and it makes the rationale legible: it exists to cover the case the subscriber misses.

2. **Expanded the inline comment** into a question + bullet list so the rationale survives a re-read:

    ```js
    // Why reflect here too, when the subscriber already calls `#reflect`?
    // - The Computed dedupes recomputes whose new value equals its
    //   cached value. `el.v = 5` when the default resolves to 5 produces
    //   no subscriber fire, so reflection from `#onComputedChange` is
    //   skipped (#105).
    // - We can't read `signal.value` to get the post-convert form: that
    //   would force a sync recompute, fire subscribers immediately, and
    //   split multi-prop dep cascades across two propchange drains.
    // So apply `spec.convert` manually and call `#reflect` directly.
    ```

The two bullets are the surprises a reader is likely to bump into: why this site exists at all (the dedupe), and why we don't just read `signal.value` for the post-convert form (cascade batching). The PR description's "Code shape" and "Implementation" sections are updated to match.

LeaVerou: Instead of jumping through hoops and redesigning API shape so we can read a signal without triggering getter side effects maybe we can just not make the internal value private? E.g. a symbol or a good ol' `_value`?

src/signals.js
```diff
@@ -187,17 +187,7 @@ export class Computed extends Signal {
...
-		tracking = null;
-		let old = super.value;
-		tracking = prev2;
-
-		if (!this.equals(value, old)) {
-			// Use Signal.prototype.value setter directly (bypasses no-op Computed setter)
-			Object.getOwnPropertyDescriptor(Signal.prototype, "value").set.call(this, value);
-		}
+		// Bypass Computed's read-only setter; Signal#set handles the equals dedupe.
+		Object.getOwnPropertyDescriptor(Signal.prototype, "value").set.call(this, value);
```
LeaVerou: This is a smell. We use hacks to work around code we don't control (third-party dependencies, browser implementations etc). We don't use hacks to work around our own code, we figure out what the actual issue is and we fix it!

src/plugins/props/util/Prop.js
```diff
@@ -271,15 +284,37 @@ let Self = class Prop {
...
+			// Why reflect here too, when the subscriber already calls `#reflect`?
+			// - The Computed dedupes recomputes whose new value equals its
+			//   cached value. `el.v = 5` when the default resolves to 5 produces
+			//   no subscriber fire, so reflection from `#onComputedChange` is
+			//   skipped (#105).
+			// - We can't read `signal.value` to get the post-convert form: that
+			//   would force a sync recompute, fire subscribers immediately, and
+			//   split multi-prop dep cascades across two propchange drains.
+			// So apply `spec.convert` manually and call `#reflect` directly.
+			if (source === "property" && parsedValue !== undefined) {
```
LeaVerou: Same comment applies here too. We seem to be working around our own code.

src/plugins/props/util/Prop.js
```diff
@@ -175,13 +183,14 @@ let Self = class Prop {
 							}
 						}
 					}
-					if (value != undefined && this.spec.convert) {
+					if (this.spec.convert) {
```
DmitrySharabin, line 184: Restore the pre-signal contract to call `convert` (if it's defined) on any value.

DmitrySharabin:
@LeaVerou, thank you for the review. Could you please give the PR another look? In the revised approach, we moved attribute reflection out of the Computed's subscriber and onto a subscriber on `rawSignal`, restoring pre-signals' “reflect only when the user wrote (or set/removed an attribute)” scope. The Computed's subscriber now only updates `element.props` and fires `propchange` (convert-dep / get-dep / default-dep cascades no longer touch the attribute, matching pre-signals exactly). As a side benefit, the inline `#reflect` workaround in `prop.set` falls away, and `Computed.#compute` can now use `super.value = value` instead of reaching into `Signal.prototype` via `Object.getOwnPropertyDescriptor` to bypass Computed's own no-op setter.

LeaVerou (approved):

Issue #104, opened 2026-04-30
[Props] static props not inherited from parent class


DmitrySharabin:
## Symptom

With a class hierarchy where both Parent and Child declare `static props`, Parent's props are never installed — no accessor descriptors on `Parent.prototype`, no `Props` instance on Parent.

```js
class Parent extends NudeElement {
  static props = { a: { type: String } };
}
class Child extends Parent {
  static props = { c: { type: String } };
}
customElements.define("x-child", Child);

let el = new Child();
el.a = "value";          // sets a data property; no parsing, no propchange
console.log(el.a);       // "value" — but Parent's accessor never ran

Object.getOwnPropertyDescriptor(Parent.prototype, "a");  // undefined

// `props` here is the internal symbol the plugin uses to attach the
// per-class Props instance — exported from `src/plugins/props/index.js`.
Parent[props];                                            // undefined
```

## Root cause

The `setup` hook in [`src/plugins/props/index.js#L26-L30`](https://github.com/nudeui/element/blob/main/src/plugins/props/index.js#L26-L30) only checks `Object.hasOwn(this, "props")`:

```js
setup () {
    if (Object.hasOwn(this, "props")) {
        this.defineProps();
    }
},
```

`hooks-common`'s `setup` cascade runs once with `this` = most-derived class only. So when `Child` is the most-derived, only `Child`'s `static props` is processed: `Child.defineProps()` is called, which lazily creates `Child[props]` and installs accessor descriptors on `Child.prototype`. Parent's `static props` is never processed — its accessor descriptors are never installed on `Parent.prototype`, and `Parent[props]` is never created.

The `// TODO how does this work if attributeChangedCallback is inherited?` at [`index.js#L9`](https://github.com/nudeui/element/blob/main/src/plugins/props/index.js#L9) suggests this territory was known to be unfinished.

## Reproducible on

`main` and `props-batched-drain`. Pre-existing — the static-only-on-most-derived semantic of `setup` predates the recent props refactors. Latent in production because no consumer in `nude-element` or `color-elements` declares `static props` on a non-leaf class (every color-element component extends `ColorElement`, which has zero props of its own).

## Suggested fix direction

Walk the class chain in `setup`. Either:

- **Local fix:** have the props `setup` hook walk ancestors via `getSupers(this, NudeElement)` and call `defineProps` for each ancestor that has its own `static props`. Each class still gets its own `Props` instance; descriptors land on the correct prototypes.
[...]

PR #109, opened 2026-05-12
Fix `setAttribute()` after mount not updating reflected props (closes #98)

### LINKED ISSUE #98: `setAttribute()` after mount does not update the property on plain `reflect` props

DmitrySharabin:
## TL;DR

Calling `el.setAttribute("v", "100")` after mount on a `reflect: true` prop did not update `el.v`, because `customElements.define` cached an empty `observedAttributes` list and a `null` `attributeChangedCallback` at registration time. This PR makes ACB always present on the prototype chain via the props plugin's `provides`, which lets the spec read `observedAttributes` and triggers the eager static getter that populates it.

## Why

Per [HTML spec §CustomElementRegistry.define](https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition):

1. For each lifecycle-callback name, `? Get(prototype, callbackName)`.
2. `observedAttributes` is initialized to an empty sequence.
3. **Only if `lifecycleCallbacks["attributeChangedCallback"]` is non-null**, `? Get(constructor, "observedAttributes")`.

Pre-this-PR, `attributeChangedCallback` was wired inside `first_constructor_static` (which fires on first instance construction, after registration). The browser snapshotted a `null` ACB at `define` time and so never read `observedAttributes` either. Pre-set attributes worked only because `Props.initializeFor` re-walks `observedAttributes` on mount.

The natural fix is to make sure ACB is reachable through the prototype chain by the time `customElements.define` runs.

### `src/plugins/props/index.js`

`attributeChangedCallback` becomes a regular entry in the plugin's `provides`. `addPlugins` lands it on `NudeElement.prototype` once, and every subclass — at any depth of inheritance — inherits it through the chain. So `Get(Class.prototype, "attributeChangedCallback")` returns it during `define`, which causes the spec to read `Class.observedAttributes`, which triggers the eager static getter and populates the cache.

Chaining to a user-declared override follows the same pattern as `connectedCallback` in [src/element/members.js](src/element/members.js): `getSuperMethod(this, provides.attributeChangedCallback)?.call(this, ...)` then the plugin's dispatch into `Props.attributeChanged`. A subclass that overrides `attributeChangedCallback` must call `super.attributeChangedCallback(...)` to compose, which is standard JS — see "Known caveat" below.

The eager static `observedAttributes` getter on `providesStatic` is kept — its only job now is to lazily call `defineProps()` and return the list. The previous ACB install block inside the getter is gone (it was the chicken-and-egg: ACB-on-prototype is the precondition for the getter ever firing).

The cache is still reserved as `[]` **before** `defineProps` runs, so a `define-props` hook that reads `Class.observedAttributes` hits the cache instead of recursing — same re-entry-guard shape as `attachShadow` in [src/plugins/shadow/index.js](src/plugins/shadow/index.js).

The props plugin's `setup` hook still skips `defineProps()` when the static getter has already cached `observedAttributes`, so plugin-contributed additive calls (e.g. `events/onprops` appending synthetic `on*` props via `this.defineProps(newProps)`) keep working after the static install.

### `test/util/FakeElement.js`

`FakeElement.with` now mirrors the spec faithfully when building its per-class `CustomElementDefinition`-shaped snapshot:

1. Read `attributeChangedCallback` from the prototype chain first.
2. **Only if non-null**, read `observedAttributes` from the constructor.

This closes the loophole the previous test harness had: it used to read `Class.observedAttributes` unconditionally, which triggered the eager install regardless of whether ACB was actually on the prototype — making the unit test pass even when the real-browser path was broken (the central complaint of the PR review). With this conditional, the unit suite is now load-bearing on the spec path.

The snapshot logic was also extracted into a new `FakeElement.define(Class)` static method, so a test can swap a class's prototype ACB and re-snapshot — modelling how `customElements.define` snapshots each registered class independently. Used by the subclass-override regression test below.

`setAttribute` / `removeAttribute` continue to dispatch through the snapshot, the same path a real browser takes via `lifecycleCallbacks`.

### `test/Props.js`

The existing `"Post-mount setAttribute updates the property"` test now exercises the real registration path end-to-end. Without `provides.attributeChangedCallback`, it fails with `Got undefined, expected 100` — the literal symptom from #98.

The `Class.defineProps()` additive-contract test (added in an earlier commit on this branch) still pins down that `defineProps({ bar: {} })` after the registration-time install must still register `bar`.

A new `Subclass attributeChangedCallback override` group pins both halves of the new contract (see "Known caveat" below):

- **Override that calls super updates the reflected prop** (expect `100`) — verifies that chaining via `super.attributeChangedCallback(...)` still flows through `Props.attributeChanged`.
- **Override that omits super leaves the reflected prop unchanged** (expect `undefined`) — verifies that forgetting `super` does silently drop the reflection, so the contract is observable in tests.

Both halves were sanity-checked: inverting either branch causes exactly the opposite test to fail with the expected symptom.

## Known caveat — behavior change vs pre-PR-109

Pre-PR-109, `first_constructor_static` unconditionally wrapped any existing ACB on the prototype chain. A subclass could declare its own `attributeChangedCallback` and **still** get the plugin's reflected-prop dispatch fired automatically, without calling super. PR-109 partially broke that auto-wrap (its `!Object.hasOwn(this.prototype, "attributeChangedCallback")` guard skipped wrapping for any subclass that had an own ACB). This PR drops the auto-wrap entirely.

**The new contract**: any subclass that overrides `attributeChangedCallback` must call `super.attributeChangedCallback(name, oldValue, value)` to compose with the plugin. The same convention already governs `connectedCallback` / `disconnectedCallback` in this codebase. Pinned by the new regression test in `test/Props.js`.

**Why auto-wrap can't be restored without an API change.** JavaScript class-method syntax installs a data property on the subclass prototype via internal `DefineOwnProperty` — that path bypasses any setter trap we could install on the base prototype, so there is no language-level way to intercept "subclass just defined an ACB" without consumer cooperation. The spec snapshots ACB at `customElements.define` time and never looks at the prototype again, so wrapping later (eager getter, first construction, etc.) is invisible to the cached callback. The pre-PR-109 wrapping landed at first construction — which is precisely why the spec snapshot was already cached (with `null`) by then, i.e. the auto-wrap *was* the bug. You can have auto-wrap or spec-correct registration, not both, without introducing a `NudeElement.define(name)` registration helper.

**Footgun severity.** This is the worst lifecycle hook to require manual super-chaining on, from a footgun-cost perspective: forgetting `super` silently re-introduces #98 for that one subclass. By contrast, forgetting `super.connectedCallback?.()` only breaks the `connected` hook (propchange queue drain on reconnect — obscure). A registration helper that would restore auto-wrap without re-introducing the spec problem is proposed for discussion in #110.

## Out of scope

- `// FIXME how to combine with existing observedAttributes?` ([src/plugins/props/index.js](src/plugins/props/index.js)) — preserved. Merging a user-declared `static observedAttributes` with derived ones is an open question; current shadowing semantics are kept.
- ACB inheritance edge cases — tracked separately under #104.
- A registration helper (`NudeElement.define(name)`) that auto-wraps lifecycle callbacks uniformly — not introduced here. Proposed for discussion in #110.

## Test plan

- [x] `npm test` — 102/103 pass (1 pre-existing skip in `split()`).
- [x] Negative control: commenting out the new `provides.attributeChangedCallback` → `"Post-mount setAttribute updates the property"` fails with `Got undefined, expected 100`, proving the prototype-resident ACB is what enables the spec path.
- [x] Negative control: removing the `acb ?` conditional in `FakeElement.with` → tests still pass but the harness stops mirroring the spec; confirms the conditional itself is load-bearing on the regression check.
- [x] Negative control on the new override-contract test: unconditionally chaining to super → `"Override that omits super leaves the reflected prop unchanged"` fails with `Got 100, expected undefined`; never chaining → `"Override that calls super updates the reflected prop"` fails with `Got undefined, expected 100`. Both halves are load-bearing.
- [x] Real-browser spot check (manual): in the deploy preview, on `<color-picker>` (which inherits two levels — `ColorPicker → ColorElement → NudeElement`), call `setAttribute` on a reflected prop from DevTools and confirm the JS property updates.

### DIFF
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -77,7 +77,7 @@ defineLazyProperty(provides, "ignoredAttributes", {
 
 const providesStatic = {
 	defineProps (def = this.props) {
-		if (def instanceof Props && def.Class === this) {
+		if ((def instanceof Props && def.Class === this) || this[props].size > 0) {
 			// Already defined
 			return null;
 		}
@@ -101,11 +101,10 @@ const providesStatic = {
 			return this[observedAttributes];
 		}
 
-		// Reserve the cache before firing setup() so re-entry from a setup hook
-		// reading Class.observedAttributes returns the in-flight list instead of
-		// recursing into setup().
+		// Reserve the cache before defineProps so a `define-props` hook that reads
+		// Class.observedAttributes returns the in-flight list instead of recursing.
 		this[observedAttributes] = [];
-		this.setup();
+		this.defineProps();
 
 		// FIXME how to combine with existing observedAttributes?
 		return (this[observedAttributes] = this[props].observedAttributes);
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -4,6 +4,7 @@ import { defineOwnProperty } from "xtensible/util";
 import { defineLazyProperty } from "../../util/lazy.js";
 
 export const { props } = symbols.known;
+const { observedAttributes } = symbols.known;
 
 function first_constructor_static () {
 	// TODO how does this work if attributeChangedCallback is inherited?
@@ -12,14 +13,6 @@ function first_constructor_static () {
 		this.constructor[props].attributeChanged(this, name, oldValue, value);
 		_attributeChangedCallback?.call(this, name, oldValue, value);
 	};
-
-	// FIXME how to combine with existing observedAttributes?
-	if (!Object.hasOwn(this, "observedAttributes")) {
-		Object.defineProperty(this, "observedAttributes", {
-			get: () => this[props].observedAttributes,
-			configurable: true,
-		});
-	}
 }
 
 const hooks = {
@@ -102,6 +95,21 @@ const providesStatic = {
 	// 		return [...superProps, ...thisProps];
 	// 	},
 	// }),
+
+	get observedAttributes () {
+		if (Object.hasOwn(this, observedAttributes)) {
+			return this[observedAttributes];
+		}
+
+		// Reserve the cache before firing setup() so re-entry from a setup hook
+		// reading Class.observedAttributes returns the in-flight list instead of
+		// recursing into setup().
+		this[observedAttributes] = [];
+		this.setup();
+
+		// FIXME how to combine with existing observedAttributes?
+		return (this[observedAttributes] = this[props].observedAttributes);
+	},
 };
 
 defineOwnProperty(providesStatic, props, function () {
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -1,6 +1,6 @@
 import Props from "./util/Props.js";
 import { symbols } from "xtensible";
-import { defineOwnProperty } from "xtensible/util";
+import { defineOwnProperty, getSuperMethod } from "xtensible/util";
 import { defineLazyProperty } from "../../util/lazy.js";
 
 export const { props } = symbols.known;
@@ -41,11 +41,15 @@ const hooks = {
 };
 
 const provides = {
-	// ...composed({
-	// 	attributeChangedCallback (name, oldValue, value) {
-	// 		this.constructor[props].attributeChanged(this, name, oldValue, value);
-	// 	},
-	// }),
+	// Must be on the prototype chain by the time customElements.define runs:
+	// the spec only reads observedAttributes if attributeChangedCallback is non-null.
+	// https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition
+	attributeChangedCallback (name, oldValue, value) {
+		// Same as super.attributeChangedCallback?.()
+		getSuperMethod(this, provides.attributeChangedCallback)?.call(this, name, oldValue, value);
+
+		this.constructor[props].attributeChanged(this, name, oldValue, value);
+	},
 };
 
 // Internal prop values
@@ -97,18 +101,6 @@ const providesStatic = {
 		// hook listener) gets the in-flight list instead of recursing.
 		this[observedAttributes] = [];
 		this.defineProps();
-
-		// Must land before customElements.define snapshots the prototype callbacks;
-		// a later patch is invisible to post-mount setAttribute reactions.
-		// https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition
-		// TODO how does this work if attributeChangedCallback is inherited?
-		if (!Object.hasOwn(this.prototype, "attributeChangedCallback")) {
-			let _attributeChangedCallback = this.prototype.attributeChangedCallback;
-			this.prototype.attributeChangedCallback = function (name, oldValue, value) {
-				this.constructor[props].attributeChanged(this, name, oldValue, value);
-				_attributeChangedCallback?.call(this, name, oldValue, value);
-			};
-		}
 
 		// FIXME how to combine with existing observedAttributes?
 		return (this[observedAttributes] = this[props].observedAttributes);
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -6,15 +6,6 @@ import { defineLazyProperty } from "../../util/lazy.js";
 export const { props } = symbols.known;
 const { observedAttributes } = symbols.known;
 
-function first_constructor_static () {
-	// TODO how does this work if attributeChangedCallback is inherited?
-	let _attributeChangedCallback = this.prototype.attributeChangedCallback;
-	this.prototype.attributeChangedCallback = function (name, oldValue, value) {
-		this.constructor[props].attributeChanged(this, name, oldValue, value);
-		_attributeChangedCallback?.call(this, name, oldValue, value);
-	};
-}
-
 const hooks = {
 	setup () {
 		// Skip if the static observedAttributes getter already ran the install —
@@ -39,8 +30,6 @@ const hooks = {
 			this.addEventListener("propschange", this.updated);
 		}
 	},
-
-	first_constructor_static,
 
 	constructed () {
 		this.constructor[props].initializeFor(this);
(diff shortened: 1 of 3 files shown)


DmitrySharabin:
> Posted on behalf of @DmitrySharabin. Investigation and writeup by Claude (Anthropic).

Empirical check in a real browser (Chromium 150) on `color-elements` (the primary consumer of nude-element) with this PR applied: the eager `observedAttributes` getter is never invoked by `customElements.define` for the registered classes. The registration-time snapshot of `attributeChangedCallback` is therefore still `null`, and post-mount `setAttribute` does not update reflected props — the very symptom #98 reports. The unit tests in this PR pass because `FakeElement.setAttribute` calls `Props.attributeChanged` directly, which bypasses the `customElements.define`-snapshot path the bug is actually about — so they certify nothing about real-browser behaviour.

## Reduction (same page where nude-element is loaded)

```js
class Direct extends NudeElement { static props = { foo: { reflect: { from: true } } } };
customElements.define("x-direct", Direct);
// ✅ eager observedAttributes getter fires; ACB patch installed; post-mount
//    setAttribute updates the prop.

class Mid extends NudeElement {};                    // empty intermediate
class Leaf extends Mid { static props = { foo: { reflect: { from: true } } } };
customElements.define("x-leaf", Leaf);
// ❌ eager getter never fires (verified with console.warn at the top of the
//    getter — silent during define). ACB patch never installed at registration.
//    Post-mount setAttribute does NOT update the prop.
```

`Mid` doesn't need `static props`, `static plugins`, or anything else — the mere presence of an intermediate class between the registered class and `NudeElement` is enough to break the lookup. Every color-element matches the bottom shape (`ColorPicker extends ColorElement extends NudeElement`), so in practice this PR doesn't fix #98 for the primary consumer.

## What I verified

- Minimal HTML page that imports only `<color-picker>` via the importmap, no docs-page scaffolding, no `html-demo`, no wrappers around `customElements.define`, no instrumentation in the loaded `props/index.js` module path.
- `console.warn` injected at the top of the eager getter is silent during page load (even after 2.5s settle).
- The class *is* registered (`customElements.get("color-picker")` returns it), so `customElements.define` *was* called.
- After registration, `Object.getOwnPropertyDescriptor(ColorPicker.prototype, "attributeChangedCallback")` is `undefined`.
- Manually reading `ColorPicker.observedAttributes` from DevTools *does* fire the getter — confirming the getter is on the chain and reachable via ECMAScript \`Get\`, just not via the path `customElements.define` takes here.

A hand-rolled minimal mimic of the same shape (3- and 4-level inheritance with `observedAttributes` installed via `Object.defineProperty` inside a `static {…}` block) does *not* reproduce. So the failure isn't generic Chromium behaviour with inherited static getters — it's something specific to nude-element's class-construction chain that I couldn't fully isolate. The failure persists with an empty intermediate class (no `static props`, no `static plugins`, no `static {…}` block) declared in the same script as the leaf class.

## Direction

For this PR to actually close #98, the eager install needs to land as an **own property** of the registered subclass before `customElements.define` runs — not via prototype-chain `Get` reaching the getter on `NudeElement`. Two shapes that work:

1. **In nude-element.** Ship a `Class.prepareForDefine()` helper (or fold into the api plugin's static side) that subclasses call right before `customElements.define`. Reading `this.observedAttributes` inside it triggers the existing eager getter, which installs the cache and the `attributeChangedCallback` patch on the *registered* subclass as own properties. The existing `setup` hook is too late — it fires from `constructed`, after registration.
2. **In consumer code.** A `define` wrapper (e.g. `ColorElement.define` in color-elements) reads `Class.observedAttributes` before calling native `customElements.define`. This is what an earlier diagnostic wrapper I added effectively did — which is why the tests looked clean before I removed the wrapper, and why I initially (incorrectly) reported the fix as verified. It's a real workaround, but the proper fix belongs here.

## Mea culpa

My implementation-notes comment on #98 called the unit-test approach "load-bearing" with a real-DOM smoke test as a complement. The complement was never actually exercised; without it, the unit suite certifies nothing about the snapshot path. A no-wrapper Playwright load of `/src/color-picker/` (or anything that registers a class via two-step inheritance from `NudeElement`) would have caught this.

DmitrySharabin:
> Posted on behalf of @DmitrySharabin. Verification by Claude (Anthropic).

Verified `a3539c5` works end-to-end in `color-elements` against the docs page (`/src/color-picker/`), fresh load in Chromium 150, no instrumentation, no `customElements.define` wrappers, no other probe trickery.

## What I checked

```
acbDescOnSubclass: false              ← ColorPicker doesn't shadow; inherits
acbOwner:          "NudeElement"      ← lands on the parent prototype, as designed
inMarkup setAttribute("space","p3"): "oklch" → "p3"  ✓
fresh   setAttribute("space","p3"): "oklch" → "p3"  ✓
```

Cross-component smoke (programmatically created, mounted, then `setAttribute`):

| Case | Result | Pass |
|---|---|---|
| `color-swatch.value` (`reflect:{from:true}`) | `""` → `"rebeccapurple"` | ✓ |
| `color-picker.color` (`reflect:{from:"color"}`) | `oklch(60% 0.12 180)` → `oklch(80% 0.08 60)` | ✓ |
| `channel-slider.value` (Number, `reflect:{from:"value"}`) | `50` → `75` | ✓ |
| `color-chart.xMin` (attr `xmin`, prop `xMin`) | `"auto"` → `"25"` | ✓ |

## Note on the approach

Putting `attributeChangedCallback` in `provides` is cleaner than the earlier eager-getter install I had floated in #98 and on this PR — it sidesteps the chicken-and-egg the spec creates ("only read `observedAttributes` if ACB is non-null on the chain"), works at any inheritance depth without per-subclass coordination, and folds chain delegation into a single `getSuperMethod` call instead of a per-subclass closure capture. The strengthened `FakeElement.with` (ACB-first lookup, observedAttributes only if non-null) now exercises the same path real browsers do, so the unit suite actually certifies the spec contract.

LGTM from my end.

Issue #113, opened 2026-05-14
Roll back signals and batching and reimplement them


LeaVerou:
#91 and #102 caused several regressions (and uncovered some existing bugs). Some of which we fixed with subsequent PRs, but generally piling fixes on top of existing broken code tends to produce worse results and higher complexity than going back, fixing the spec, and reimplementing.

So I think we need to bite the bullet and do just that. 

What makes the rollback tricky is that once we discovered the regressions, we added tests for them. These need to stay. FakeElement too (or use a library that does it properly). But simply keeping all the tests from the rewrite is a no-go, since many depend on internal implementation details of the rewrite (a problem in itself).
Also, there are pre-existing bugs we discovered and fixed since then.

I propose this plan:
- We roll back to e344b1f68bb51d3d0aa0ec9fb8065fead57fe215 (last commit before signals)
- Let's create a list of requirements here for behavior we want to preserve and ask Claude to create tests for them
- Fix any pre-existing bugs, and add tests to prevent regressions
- Then ask Claude to implement coalescing **first** (#51) then signals (#79). I'll edit the issue descriptions to include more details, so we can point Claude to the issues directly.

## Bugs to fix before implementing signals

- #98 
- #100
- #104 
- #106

We should also add tests for these behaviors.

## Behavioral baseline (tests to add before reimplementing)

First, we should find a library instead of implementing FakeElement ourselves and add it as a dev dependency. This cannot be a new problem.

These should all pass on e344b1f. They're the regression net protecting consumer-observable behavior through the signals refactor.

**Basic API** (mostly already present in `test/Prop.js` constructor() group — keep what's there, expand only where gaps exist):

- Construction normalizes `type`, `default`, `reflect`, etc.
- `reflect: true | false | "name" | { from, to }` parses correctly into `fromAttribute` / `toAttribute`.
- `Props.observedAttributes` derives from reflecting props.
- `Props.add(name, spec)` and `Props.add({...})` both work.

**Sync IDL semantics** to prevent #111 

- `el.x = v; el.x` returns `v` synchronously.
- `el.x = 1; el.x = 2; el.x = undefined` fires three `propchange` events in order, synchronously
- After `el.x = v`, reading any prop derived from `x` (via `get`, `convert`, or function `default`) returns the new derived value on the same turn.
- After `el.setAttribute("x", "foo")`, `el.x === "foo"` synchronously.

**Defaults**:

- Plain literal default returns on read when nothing was written.
[...]

PR #114, opened 2026-05-14
Behavioral baseline test suite for props/events plugins (#113)

DmitrySharabin:
## Summary

A happy-dom-based behavioral test suite for the props and events plugins, serving as the issue #113 regression net (rolling back and reimplementing signals).

- Adopts main's behavioral tests, regrouped under `test/plugins/props/` and `test/plugins/events/`; skips ones that test signals internals.
- Covers the full #113 behavioral baseline and the four "bugs to fix" (#98, #100, #104, #106).
- Tests capture the value chain through each action — a regression in event payload, default resolution, or derived recomputation is caught right away.
- Some tests fail by design — they surface the pre-existing bugs #113 is meant to fix, kept visible rather than skipped.

### DIFF
--- a/test/plugins/props/computed.js
+++ b/test/plugins/props/computed.js
@@ -1,22 +1,15 @@
 export default {
 	name: "Computed properties",
 
-	run ({ actions = [], read }) {
-		if (actions.length === 0) {
-			return this.data.element[read];
-		}
-
-		let stages = [this.data.element[read]];
-		for (let action of actions) {
-			action(this.data.element);
-			stages.push(this.data.element[read]);
-		}
-		return stages;
-	},
-
 	tests: [
 		{
 			name: "get() updates when its dependency changes",
+			run () {
+				let { element } = this.data;
+				let before = element.derived;
+				element.base = 5;
+				return [before, element.derived];
+			},
 			arg: {
 				props: {
 					base: { type: Number, default: 1 },
@@ -26,13 +19,17 @@ export default {
 						},
 					},
 				},
-				actions: [el => (el.base = 5)],
-				read: "derived",
 			},
 			expect: [2, 10],
 		},
 		{
 			name: "spec.equals: a tolerated dependency change leaves the cached value",
+			run () {
+				let { element } = this.data;
+				let before = element.derived;
+				element.base = 42.05;
+				return [before, element.derived];
+			},
 			arg: {
 				props: {
 					base: { type: Number, default: 42 },
@@ -43,8 +40,6 @@ export default {
 						equals: (a, b) => Math.abs(a - b) < 0.1,
 					},
 				},
-				actions: [el => (el.base = 42.05)],
-				read: "derived",
 			},
 			expect: [42, 42],
 		},
--- a/test/plugins/props/computed.js
+++ b/test/plugins/props/computed.js
@@ -2,11 +2,16 @@ export default {
 	name: "Computed properties",
 
 	run ({ actions = [], read }) {
-		for (let action of actions) {
-			action(this.data.element);
+		if (actions.length === 0) {
+			return this.data.element[read];
 		}
 
-		return this.data.element[read];
+		let stages = [this.data.element[read]];
+		for (let action of actions) {
+			action(this.data.element);
+			stages.push(this.data.element[read]);
+		}
+		return stages;
 	},
 
 	tests: [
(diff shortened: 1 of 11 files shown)


test/plugins/props/computed.js
```diff
@@ -0,0 +1,52 @@
...
+			arg: {
+				props: {
+					base: { type: Number, default: 1 },
+					derived: {
+						get () {
+							return this.base * 2;
+						},
+					},
+				},
+				actions: [el => (el.base = 5)],
```
LeaVerou: This is a smell. The declarativeness of hTest allows us to avoid repetitive code. It doesn't mean we have to bend over backwards to turn imperative code into declarative code. `actions` is inherently imperative. 

The rule of thumb is: what abstraction makes the tests easier to understand for humans?

This seems like the kind of abstraction that introduces indirection without improving either maintainability OR clarity.

LeaVerou: @DmitrySharabin if the hTest skill produces code like this, we need to fix the skill.

test/plugins/props/defaults.js
```diff
@@ -0,0 +1,121 @@
...
+			arg: {
+				props: {
+					src: { type: String, default: "initial" },
+					mirror: { type: String, defaultProp: "src" },
+				},
+				actions: [el => (el.mirror = "explicit"), el => (el.src = "after-sever")],
+				read: "mirror",
+			},
+			expect: ["initial", "explicit", "explicit"],
+		},
```
LeaVerou, line 31: I think a good hTest principle is "write the tests _first_, in the way you'd find your intent easier to express with maximum signal-to-noise, and write a `run()` (or `beforeEach()` etc) **at the end**".

E.g. trying this here:

```suggestion
		{
			name: "defaultProp severs on explicit write",
			arg: {
				props: {
					src: { type: String, default: "initial" },
					mirror: { type: String, defaultProp: "src" },
				},
			},
			run (el) {
				el.mirror = "explicit";
				el.src = "after-sever";
				return el.mirror;
			}
			expect: ["initial", "explicit", "explicit"],
		},
```

I find this much easier to read. Which hints that the right primitive is not a root-level `run()`, but a root-level `beforeEach()`.

test/plugins/props/lifecycle.js
```diff
@@ -0,0 +1,30 @@
+export default {
+	name: "Disconnect / reconnect lifecycle",
+
+	tests: [
+		{
+			name: "Queued propchange events drain on reconnect",
+			run () {
+				let { element } = this.data;
```
LeaVerou, line 8: Just making sure that this is different per test? (I think so, but not 100% sure)

test/plugins/props/propchange.js
```diff
@@ -0,0 +1,190 @@
...
+				props: {
+					src: { type: String, default: "initial" },
+					mirror: { type: String, defaultProp: "src" },
+				},
+			},
+			expect: [
+				["src", "initial"],
+				["mirror", "initial"],
+				["src", "changed"],
+				["mirror", "changed"],
```
LeaVerou, line 21: Shouldn't this be a dictionary?

test/plugins/props/reflection.js
```diff
@@ -3,22 +3,124 @@ export default {
...
+				return this.data.element.bar;
+			},
+			arg: {
+				props: { bar: { type: Number, reflect: true } },
+				attributes: { bar: "42" },
+			},
+			expect: 42,
+		},
+		{
+			name: "Default is not reflected to the attribute on mount",
```
LeaVerou: I would consider this a `defaults.js` test (though really it's an integration test between the two features…)

LeaVerou (approved):

test/plugins/props/lifecycle.js
DmitrySharabin: It is. We create it in the `beforeEach()` hook.

test/plugins/props/propchange.js
DmitrySharabin: What do you mean?

test/plugins/props/reflection.js
DmitrySharabin: Moved

test/plugins/props/propchange.js
LeaVerou: This feels like it should be an object literal (where emitted too). Anyway it doesn't matter much

Issue #110, opened 2026-05-13
RFC: registration helper to auto-wrap lifecycle callbacks (follow-up to #98 / #109)


DmitrySharabin:
## Context

This is a follow-up to the discussion on the proposed fix for #98 (PR #109, **not yet merged** and possibly still subject to change). If the approach in #109 lands roughly as-is, it shifts where `attributeChangedCallback` lives:

- **Before**: ACB was installed inside `first_constructor_static`, which fires on first instance construction. The hook unconditionally wrapped any existing ACB on the prototype chain, so a subclass that declared its own `attributeChangedCallback` still got the plugin's reflected-prop dispatch fired automatically.
- **Proposed in #109**: ACB moves to the props plugin's `provides`, so it lives on `NudeElement.prototype` and is reachable through the inherited prototype chain at `customElements.define` time. This is unavoidable for spec correctness — [`customElements.define`](https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition) reads ACB from the prototype chain at registration, before any first-instance logic runs, and uses it to gate the `observedAttributes` read.

The side effect of the shift: the previous unconditional auto-wrap goes away. A subclass that overrides `attributeChangedCallback` would have to call `super.attributeChangedCallback(...)` to compose with the plugin's reflected-prop dispatch — otherwise post-mount `setAttribute` would silently stop updating reflected props for that subclass (the exact #98 symptom, isolated to the offending subclass).

This is the worst lifecycle hook to require manual super-chaining. Other hooks (`connectedCallback` / `disconnectedCallback`) already have the same explicit-super expectation in this codebase, but their failure modes are less visible — e.g., a `connectedCallback` override that forgets super means the `connected` hook doesn't fire, which mostly affects the propchange queue drain on reconnect.

This issue exists to discuss whether a registration helper is the right way to restore auto-wrap without re-introducing the spec-correctness problem.

## Proposal

Add a static `define()` method to `NudeElement` that subclasses would call instead of `customElements.define(name, Class)`. It would wrap any own-property lifecycle callbacks on the subclass prototype so they compose with the inherited plugin methods, then delegate to native `customElements.define`.

```js
class MyElement extends NudeElement {
    static props = { v: { type: Number, reflect: true } };

    attributeChangedCallback (name, oldValue, value) {
        console.log("custom logic — no super needed");
    }
}

MyElement.define("my-element");
// vs. customElements.define("my-element", MyElement)
// which would still work but lose the auto-wrap.
```

[...]

PR #117, opened 2026-05-15
Fix `setAttribute()` after mount not updating reflected props (closes #98)

### LINKED ISSUE #98: `setAttribute()` after mount does not update the property on plain `reflect` props

DmitrySharabin:
Closes #98 — `setAttribute()` after mount didn't update reflected props because `attributeChangedCallback` wasn't on the prototype at `customElements.define` time, so the spec never read `observedAttributes`.

Per [Lea's review](https://github.com/nudeui/element/pull/117#issuecomment-4467916069): `NudeElement` now owns `attributeChangedCallback` (in `src/element/members.js`, alongside `connectedCallback` / `disconnectedCallback`) and fires an `attribute-changed` hook with `{ name, oldValue, value }`. The props plugin listens to that hook instead of overriding ACB — same shape it already uses for `connected` / `disconnected`.

**Bonus**: also flips "removeAttribute collapses a reflected prop to its default" green — same root cause.

## Test plan

- [x] `npm run test` — both regression tests pass: "Post-mount setAttribute updates the property (issue #98)" and "removeAttribute collapses a reflected prop to its default"
- [x] No regressions in unrelated tests

### DIFF
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -1,6 +1,6 @@
 import Props from "./util/Props.js";
 import { symbols } from "xtensible";
-import { defineOwnProperty, getSuperMethod } from "xtensible/util";
+import { defineOwnProperty } from "xtensible/util";
 import { defineLazyProperty } from "../../util/lazy.js";
 
 export const { props } = symbols.known;
@@ -32,19 +32,13 @@ const hooks = {
 	disconnected () {
 		this.constructor[props].pauseEvents(this);
 	},
-};
 
-const provides = {
-	// Must be on the prototype chain by the time customElements.define runs:
-	// the spec only reads observedAttributes if attributeChangedCallback is non-null.
-	// https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition
-	attributeChangedCallback (name, oldValue, value) {
-		// Same as super.attributeChangedCallback?.()
-		getSuperMethod(this, provides.attributeChangedCallback)?.call(this, name, oldValue, value);
-
+	"attribute-changed" ({ name, oldValue, value }) {
 		this.constructor[props].attributeChanged(this, name, oldValue, value);
 	},
 };
+
+const provides = {};
 
 // Internal prop values
 defineLazyProperty(provides, "props", {
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -1,30 +1,16 @@
 import Props from "./util/Props.js";
 import { symbols } from "xtensible";
-import { defineOwnProperty } from "xtensible/util";
+import { defineOwnProperty, getSuperMethod } from "xtensible/util";
 import { defineLazyProperty } from "../../util/lazy.js";
 
 export const { props } = symbols.known;
-
-function first_constructor_static () {
-	// TODO how does this work if attributeChangedCallback is inherited?
-	let _attributeChangedCallback = this.prototype.attributeChangedCallback;
-	this.prototype.attributeChangedCallback = function (name, oldValue, value) {
-		this.constructor[props].attributeChanged(this, name, oldValue, value);
-		_attributeChangedCallback?.call(this, name, oldValue, value);
-	};
-
-	// FIXME how to combine with existing observedAttributes?
-	if (!Object.hasOwn(this, "observedAttributes")) {
-		Object.defineProperty(this, "observedAttributes", {
-			get: () => this[props].observedAttributes,
-			configurable: true,
-		});
-	}
-}
+const { observedAttributes } = symbols.known;
 
 const hooks = {
 	setup () {
-		if (Object.hasOwn(this, "props")) {
+		// Skip if the static observedAttributes getter already ran the install —
+		// it's the registration-time path that fires before any instance exists.
+		if (Object.hasOwn(this, "props") && !Object.hasOwn(this, observedAttributes)) {
 			this.defineProps();
 		}
 	},
@@ -34,8 +20,6 @@ const hooks = {
 			this.addEventListener("propchange", this.propChangedCallback);
 		}
 	},
-
-	first_constructor_static,
 
 	constructed () {
 		this.constructor[props].initializeFor(this);
@@ -51,11 +35,15 @@ const hooks = {
 };
 
 const provides = {
-	// ...composed({
-	// 	attributeChangedCallback (name, oldValue, value) {
-	// 		this.constructor[props].attributeChanged(this, name, oldValue, value);
-	// 	},
-	// }),
+	// Must be on the prototype chain by the time customElements.define runs:
+	// the spec only reads observedAttributes if attributeChangedCallback is non-null.
+	// https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition
+	attributeChangedCallback (name, oldValue, value) {
+		// Same as super.attributeChangedCallback?.()
+		getSuperMethod(this, provides.attributeChangedCallback)?.call(this, name, oldValue, value);
+
+		this.constructor[props].attributeChanged(this, name, oldValue, value);
+	},
 };
 
 // Internal prop values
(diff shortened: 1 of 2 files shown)


src/plugins/props/index.js
```diff
@@ -35,19 +21,21 @@ const hooks = {
...
 const provides = {
-	// ...composed({
-	// 	attributeChangedCallback (name, oldValue, value) {
-	// 		this.constructor[props].attributeChanged(this, name, oldValue, value);
-	// 	},
-	// }),
+	// Must be on the prototype chain by the time customElements.define runs:
+	// the spec only reads observedAttributes if attributeChangedCallback is non-null.
+	// https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition
+	attributeChangedCallback (name, oldValue, value) {
```
LeaVerou: What happens when this plugin is used on an element class with its own `attributeChangedCallback`?

src/plugins/props/index.js
```diff
@@ -88,6 +76,21 @@ const providesStatic = {
...
 	// }),
+
+	get observedAttributes () {
+		if (Object.hasOwn(this, observedAttributes)) {
+			return this[observedAttributes];
+		}
+
+		// Reserve the cache before defineProps so any consumer that reads
+		// Class.observedAttributes during the install (e.g., a define-props
+		// hook listener) gets the in-flight list instead of recursing.
```
LeaVerou, line 89: How could that happen? There's nothing async here, so nothing will be executed between L88 and L89 (though I haven't checked defineProps, maybe there are async parts)

Also, this doesn't take parent classes into account at all (but maybe we're fixing that later?)

src/plugins/props/index.js
DmitrySharabin: It should call `super.attributeChangedCallback` to make it work correctly. It's a trade-off of the change.

More on this here: https://github.com/nudeui/element/issues/110

src/plugins/props/index.js
DmitrySharabin: > How could that happen? There's nothing async here

It's synchronous re-entry, not async. `defineProps()` synchronously fires `this.$hook("define-props", env)` ([props/index.js:71](https://github.com/nudeui/element/blob/main/src/plugins/props/index.js#L71)), and hook listeners run while we're still inside the getter. A hypothetical plugin that hooks `define-props` and peeks at the attr list before contributing its own:

```js
const hooks = {
  define_props (env) {
    if (this.observedAttributes.includes("foo")) { /* … */ }
  },
};
```

Without the reservation:

1. Outer read → getter, cache empty, no `Object.hasOwn` hit.
2. Getter calls `defineProps()`.
3. `defineProps` fires the `define-props` hook.
4. The listener above reads `Class.observedAttributes`.
5. Getter fires again. Cache is *still* empty (the assignment is *after* `defineProps`).
6. Step 5 → step 2 → step 3 → step 4 → step 5 …
7. 💥 stack overflow.

With the reservation, step 4's inner read hits `Object.hasOwn` and returns the in-flight `[]` immediately — same re-entry-guard shape as `attachShadow` in [src/plugins/shadow/index.js](https://github.com/nudeui/element/blob/main/src/plugins/shadow/index.js#L13-L37).

Today no plugin actually hooks `define-props`, so the guard is purely defensive — but it's a public extension point and a third-party listener would legitimately reach for `Class.observedAttributes`.

> doesn't take parent classes into account at all (but maybe we're fixing that later?)

Right — that's the `// FIXME how to combine with existing observedAttributes?` on L95. Out of scope for #117; merging derived attrs with user-declared/inherited `static observedAttributes` is the open question from #109.

src/plugins/props/index.js
LeaVerou: That won't work. Try it

src/plugins/props/index.js
LeaVerou: It's questionable whether an empty array is better in that case, but it's not a blocking issue

DmitrySharabin:
After digging into Lea's multi-level inheritance question, I think this PR is fixing the wrong layer.

**Lea's pushback holds.** Subclass-calls-super works for one level but stack-overflows at depth ≥ 2. Reproduced against the current `provides.attributeChangedCallback`: two subclasses both override ACB and call super → `RangeError: Maximum call stack size exceeded`. The same shape applies to `connectedCallback`, `disconnectedCallback`, `attachShadow`, `attachInternals` — every `provides`-installed method that uses `getSuperMethod(this, provides.foo)`. The bug isn't unique to ACB; this PR just makes a latent issue visible by adding ACB to the list.

**The root cause is structural.** User subclasses chain via native `super` (lexical, homeobject-anchored). Plugin `provides` methods can't use native `super` — object literals have no useful homeobject — so they reach for `getSuperMethod`, which is dynamically anchored on `this.constructor.prototype` and works correctly only when no subclass also overrides the same method. `getSuperMethod` is correct for its actual purpose (dynamic mixin composition — chains that aren't known in advance, where home-binding genuinely can't help). It just isn't the right tool for these specific call sites, where the method's install home is known statically.

**The right fix is [#82](https://github.com/nudeui/element/issues/82).** That issue tracks exactly this pain ("doesn't play nicely with inheritance"). The prototype already exists at `src/util/unused/composed.js`, and the `// ...composed({ attributeChangedCallback ... })` placeholder this PR removes was a sketch of the intended API. Finishing `composed()`/`constituents` gives plugin authors a declarative idiom for "method installed on a known prototype, compose with same-named methods up and down the chain." `getSuperMethod` stays available for the dynamic scenarios it was designed for — the two coexist.

**Recommendation: close this PR and use #98 as the forcing function for #82.** Any tactical patch here — home-anchored helper, drop the super call, registration helper — is throw-away work that gets deleted (or worse, entrenched) when `composed()` lands. The "subclass must call super" contract specifically has a footgun that doesn't compose at depth; landing it codifies a pattern we'd then have to migrate consumers off.

Concrete path:
1. Settle #82's design — composition order, base-value semantics, interaction with native class methods, and static composition (which also closes [#104](https://github.com/nudeui/element/issues/104) for `observedAttributes`).
2. Implement `composed()` in xtensible.
3. One nude-element PR migrates the five `provides`-method call sites — closes #98 / #82, possibly obviates [#110](https://github.com/nudeui/element/issues/110).

More work than this PR. But #98 has been open since April; using it to drive the architectural fix is probably the right tradeoff vs. shipping a workaround we'd need to undo.

🤖 Drafted by Claude Code from a full investigation in this session.

LeaVerou:
I think overridding lifecycle methods is not the answer in general. Nude Element should just call hooks from them, like it does for connected/disconnected etc. ACB is trickier because it has arguments, but not by a lot: we can just pass the arguments to the `attribute-changed` hook.

DmitrySharabin:
> I think overridding lifecycle methods is not the answer in general. Nude Element should just call hooks from them, like it does for connected/disconnected etc. ACB is trickier because it has arguments, but not by a lot: we can just pass the arguments to the `attribute-changed` hook.

Done in 55c67b9. `attributeChangedCallback` moved into `src/element/members.js` next to `connectedCallback` / `disconnectedCallback`; it fires an `attribute-changed` hook with `{ name, oldValue, value }`. The props plugin now listens via that hook instead of providing its own ACB — same pattern as `connected` / `disconnected`.

LeaVerou (approved):

PR #118, opened 2026-05-15
Apply convert to defaults and preserve explicit null writes

DmitrySharabin:
## Summary

Two related defaults bugs in `Prop.js`:

- A resolved default skipped `spec.convert`, so `{ default: 5, convert: v => v * 2 }` read as `5` instead of `10`. `get()` now applies convert after parse, matching what a user write goes through.
- Explicit `el.v = null` was overridden by the default. `get()` no longer treats `null` the same as `undefined`. To keep `removeAttribute` resetting to default, `set()` collapses null-from-attribute to `undefined`.

Unlocks the two pinned tests in the Defaults group (`Default passes through convert`, `null is preserved on a prop with a default`); suite goes 93 → 95 passing.

## Test plan

- [x] `npm test` — Defaults group fully green
- [x] Confirm reflection.js `removeAttribute collapses a reflected prop to its default` still passes

### DIFF
--- a/src/plugins/props/util/Prop.js
+++ b/src/plugins/props/util/Prop.js
@@ -156,7 +156,7 @@ let Self = class Prop {
 			value = element.props[this.name];
 		}
 
-		if (value === undefined || value === null) {
+		if (value === undefined) {
 			if (this.default !== undefined) {
 				if (this.defaultProp) {
 					value = this.defaultProp.get(element);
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):

DmitrySharabin:
Note: this PR was auto-merged by GitHub as a side effect of a restack error — the head branch was force-pushed such that it became an ancestor of the base, which GitHub interprets as "already merged." The original commit `522905e` was preserved by cherry-picking it onto `rollback-signals` as commit [`a599e8f`](https://github.com/nudeui/element/commit/a599e8f), so the change is correctly landed there.

PR #120, opened 2026-05-15
Drain queued propchange events on reconnect

DmitrySharabin:
## Summary

Replaces the implicit `isConnected` queue check with an explicit `pauseEvents` / `resumeEvents` primitive on `Props`, gated by a `#paused` WeakSet. Lifecycle hooks wire `disconnected → pauseEvents` and `connected → resumeEvents` — same fix, first-class primitive for #51 batching to build on.

Addresses [Lea's review](https://github.com/nudeui/element/pull/120#pullrequestreview-4301101960).

## Tests

96/102 PASS locally (4 pre-existing skipped). The original reconnect regression test is unchanged. One new test covers the primitive directly: pause → writes → resume, asserting no dispatch during pause and ordered drain on resume.

The 2 reds (`Post-mount setAttribute…`, `removeAttribute collapses…`) are owned by #117 and will go green once it merges — this PR deliberately doesn't carry that commit.

### DIFF
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -42,7 +42,11 @@ const hooks = {
 	},
 
 	connected () {
-		this.constructor[props].dispatchQueuedEvents(this);
+		this.constructor[props].resumeEvents(this);
+	},
+
+	disconnected () {
+		this.constructor[props].pauseEvents(this);
 	},
 };
 
--- a/src/plugins/props/index.js
+++ b/src/plugins/props/index.js
@@ -42,7 +42,7 @@ const hooks = {
 	},
 
 	connected () {
-		this.constructor[props].drainFor(this);
+		this.constructor[props].dispatchQueuedEvents(this);
 	},
 };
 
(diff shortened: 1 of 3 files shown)


LeaVerou (changes requested):
"drain" reads more like an implementation detail than a user-facing concept. Plus, drain what? It's missing context. 

I think it would be useful to make these first-class concepts, with a method to stop dispatching events and another one to resume and drain, and frame around that. I suspect this could also be more efficient, though I haven't seen the code that adds to queue to verify. And I can certainly see that being useful in other situations as well.

DmitrySharabin:
Done — reframed as `pauseEvents` / `resumeEvents` on `Props`, gated by a `#paused` WeakSet. Lifecycle hooks call them directly, so reconnect still drains and the same primitive is ready for #51 batching to build on. Coalescing is intentionally out of scope here to keep the PR minimal.

LeaVerou (approved):
We should coalesce, but this can be done in a separate PR

PR #121, opened 2026-05-16
Make happy-dom Node-only via conditional `#dom-polyfill` import

DmitrySharabin:
Routes happy-dom through a `package.json` `”imports”` condition so the test files load the polyfill in Node and resolve `#dom-polyfill` to a no-op in browsers. Each DOM-using test file declares the dependency explicitly with `import "#dom-polyfill";` — no TLA, no boot wrapper, no `index-fn.js` split.

### How it works

- `package.json` `"imports": { "#dom-polyfill": { "node": “./test/util/dom-polyfill.js” } }` — Node resolves the bare specifier to the polyfill module; the `"node"` condition is the only one defined, so other runtimes never load it.
- [`test/util/dom-polyfill.js`](test/util/dom-polyfill.js) — synchronous: registers happy-dom, restores native `Event`/`EventTarget` (hTest's classes loaded before it and captured the natives), and defers restoring native `CustomEvent` to a microtask so `PropChangeEvent extends CustomEvent` in `src/` captures happy-dom's at module-load time but hTest's runtime `new CustomEvent` gets native.
- Browsers use their import map (e.g., `"#dom-polyfill": "data:text/javascript,"`) — see #123 for the browser harness.

### Verified

- Node: `npx htest test/index.js` → 98/102 pass, 4 skipped (baseline).
- Each affected test file also runs standalone in Node (e.g. `npx htest test/Prop.js`) without needing `test/index.js` as a wrapper — addresses the “tests must be independent” review on the earlier shape of this PR.

[A separate comment below](https://github.com/nudeui/element/pull/121#issuecomment-4491866855) recaps the alternatives we tried (linkedom, jsdom, `@lit-labs/ssr-dom-shim`, `node --import`) and why none of them worked.

**P.S.** To make the tests fully independent, we need to move (duplicate) the `beforeEach()` and `afterEach()` lifecycle hooks (where we create and delete elements for tests) to every leaf file that depends on them. We probably should do it anyway (at least to be able to test NudeElement features in the browser where hTest doesn't support nested tests), but I'd suggest doing that in a follow-up PR.


LeaVerou (changes requested):
Instead of turning all our imports into async imports, which doesn't scale, we can just import (statically) a file which checks if we have a DOM and if not, dynamic imports happy-dom and does the other stuff. Then all the complexity lives in that file, which is imported normally before other imports in every test that needs it. What am I missing?

DmitrySharabin:
Tried that first (it was actually our initial shape). The literal version — `test/index.js` keeps static imports, the conditional `await import("@happy-dom/global-registrator")` lives inside `test/util/happy-dom.js` — fails on Node 24 with `Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'`. ESM lets sibling modules evaluate while a TLA-holding sibling is paused, so `PropChangeEvent` (loaded transitively via `Prop.js`) does `extends CustomEvent` before happy-dom registers and ends up extending native `CustomEvent`.

f9e3a29 keeps the spirit of your suggestion: `test/index.js` is a thin boot wrapper that statically imports `happy-dom.js` (whose TLA resolves before the body runs), then dynamic-imports `test/index-fn.js` which holds plain static imports of the test tree. Same `index.js`/`index-fn.js` convention as `src/`.

— answered by Claude (Opus 4.7) on behalf of @DmitrySharabin

LeaVerou (changes requested):
The thing is, tests need to be able to run independently, they shouldn't break if not loaded from the root.

Until we find a better solution I'd suggest loading it in every case and just not registering in browser environments. It's wasteful, but at least the wart then lives with that file, rather than infecting the entire testsuite.

LeaVerou:
> Tried that first (it was actually our initial shape). The literal version — `test/index.js` keeps static imports, the conditional `await import("@happy-dom/global-registrator")` lives inside `test/util/happy-dom.js` — fails on Node 24 with `Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'`. ESM lets sibling modules evaluate while a TLA-holding sibling is paused, so `PropChangeEvent` (loaded transitively via `Prop.js`) does `extends CustomEvent` before happy-dom registers and ends up extending native `CustomEvent`.

Why do we actually need a custom `CustomEvent` if NodeJS provides one? If additional functionality exists in browsers, why does happy-dom replace it instead of augment it?

DmitrySharabin:
## What we tried before landing on `#dom-polyfill`

The original problem on this PR — happy-dom replacing native `EventTarget`/`Event`/`CustomEvent` for hTest, plus a TLA cascade if we tried to dynamic-import it conditionally — pushed us to look for a different polyfill. None of the alternatives worked. Recap of what we tested, in case it's useful later:

1. **`@lit-labs/ssr-dom-shim`** — non-invasive (named exports, no global side-effects), but it's a registry + base classes for SSR, not a live DOM. `document.createElement()` literally returns `{}`, no `connectedCallback`/`attributeChangedCallback` lifecycle, no real `document.body`. Not a fit.
2. **jsdom** — full DOM, custom-element upgrade and lifecycle all work. But still replaces native `Event`/`EventTarget` (any spec-compliant DOM lib has to ship its own hierarchy for shadow DOM, composedPath, etc.), so it inherits the same hTest-coexistence problem.
3. **linkedom** — pure JS, browser-safe to even *load* in browsers (no Node API imports). Custom-element lifecycle, attribute observation, event bubbling all worked in our probes. But again replaces native `EventTarget` semantics; doesn't shortcut anything.
4. **Static `linkedom` + runtime `if (typeof HTMLElement === "undefined")` guard** — works in Node, but ships the full polyfill to the browser as dead weight (~50 KB) and still needs a bundler/import-map for the bare specifier.
5. **`node --import` flag** — lets the polyfill load before the main script, no test-file change needed. But the `CustomEvent` restoration microtask drains *between* `--import` and the main script, so `PropChangeEvent extends CustomEvent` in src/ captures native and dispatching it on happy-dom elements throws. Mixing `--import` with the in-file approach was worse: ESM caches the polyfill module on first load, so subsequent `import "#dom-polyfill"` is a cache hit and the inline timing trick is defeated.

The reason **none** of the dynamic-import shapes work is structural to ESM: a top-level await inside a statically-imported sibling lets the rest of the test graph evaluate during the wait. That's the timing race in #121's original commits.

### Why the `package.json` `imports` approach in this PR

Conditional loading happens at the **resolution layer**, not the JS layer. `import "#dom-polyfill"` is a plain static import — synchronous. Node's `"imports"` field with a `"node"` condition resolves it to the polyfill; in browsers the harness's import map points it at an empty `data:` URL so it's a no-op. The polyfill module loads *inline* with the test module's evaluation, so the `queueMicrotask`-deferred `CustomEvent` restoration fires *after* `PropChangeEvent extends CustomEvent` has captured happy-dom's reference but *before* hTest dispatches anything. No TLA, no race, single mechanism.

The browser test runner that consumes this — `npm run test:browser` — is in #123 (stacked on this branch).

test/util/dom-polyfill.js
```diff
@@ -0,0 +1,31 @@
...
+
+// CustomEvent is split-use. PropChangeEvent (src/) does `extends CustomEvent`
+// at module-load and is dispatched on happy-dom elements — must extend
+// happy-dom's. hTest later constructs `new CustomEvent` and dispatches on its
+// own native EventTargets — needs native. The microtask defers the swap past
+// synchronous module evaluation: by the time it fires, every `extends
+// CustomEvent` has captured happy-dom's reference; hTest's runtime calls
+// come later.
+queueMicrotask(() => {
+	globalThis.CustomEvent = NativeCustomEvent;
```
LeaVerou, line 30: Oof, this is extremely hacky :( What if we later load something that reads the DOM version of `CustomEvent` async?

LeaVerou (commented):
That's very clever. Didn't know that `import '#foo'` is a no-op in browsers! 
That said, relying on timing effects about when to override what seems *extremely* fragile. :(   

I think at this point we've sunk so much time into this that we would have been much better off using our earlier fake element approach. I would suggest reverting to that, implementing a shim, over what is not available with feature detection only. What do you think?

DmitrySharabin:
> I think at this point we've sunk so much time into this that we would have been much better off using our earlier fake element approach. I would suggest reverting to that, implementing a shim, over what is not available with feature detection only. What do you think?

Agreed. We just need to make it smarter than it was before to catch the regressions for real. I'm down for it.

DmitrySharabin:
Close it in favor of our own `FakeElement`.

LeaVerou:
> Close it in favor of our own `FakeElement`.

Let's not call it `FakeElement` though. Remember: polyfill-style.

PR #124, opened 2026-05-20
Split per-element prop state into ElementProp / ElementProps

LeaVerou:
## Summary

Splits `Prop` / `Props` (class-level metadata) from new `ElementProp` / `ElementProps` (per-element state and behavior). `element.props` becomes an `ElementProps` instance — a `Map<string, ElementProp>` — installed by the `constructed` hook.

- `Prop` exposes normalized fields (`reflect.{from,to}`, `dependencies`, `type`) and direct refs to hook fns. Descriptor captures the `Prop` directly so inherited prop accessors keep working.
- `ElementProp` self-registers in its `ElementProps` and runs shadow recovery / attr-read / default-fire inline in its constructor. Default-fire is gated on `!defaultProp && !spec.get` so computed and `defaultProp`'d props don't double-fire — they get their initial value via the cascade from their dependencies.
- `ElementProps` self-installs on `element.props` first thing in its constructor, then materializes wrappers via idempotent `forSpec(spec)`.
- Pre-mount property writes install a shadowing data property which the `ElementProp` constructor picks up via `Object.hasOwn`; pre-mount attribute writes are read by the same constructor's attr branch.

## Behavior changes

- `event.detail.element` removed (use `event.target`).
- `event.prop` is now the `ElementProp` instance (was the class-level `Prop`).
- `element.props` is the `ElementProps` Map (was a plain `{}` bag).
- `element.ignoredAttributes` moves into `ElementProps` as an internal field.

## Test plan
- [x] `npm test` — 96/100 pass, 4 skipped (parity with baseline)

### DIFF
--- a/src/plugins/props/util/ElementProp.js
+++ b/src/plugins/props/util/ElementProp.js
@@ -1,98 +1,122 @@
 import { resolveValue } from "../util.js";
 
+/**
+ * Per-element wrapper around a {@link Prop} spec.
+ * Holds the current and previous value for one prop on one element, and the
+ * per-element behavior (get, set, update, cascade) for that prop.
+ */
 export default class ElementProp {
 	/**
-	 * Internal value
+	 * Current stored value. `undefined` means no explicit value has been set,
+	 * which triggers default resolution on read.
 	 */
 	value;
 
 	/**
-	 *
-	 * @param {HTMLElement} element
-	 * @param {Prop} spec
+	 * Last value that was different from the current one (i.e. the value
+	 * before the most recent change that passed the equality check).
 	 */
-	constructor (element, spec) {
-		this.element = element;
+	oldValue;
+
+	/**
+	 * @param {ElementProps} props The owning per-element collection.
+	 * @param {Prop} spec The class-level prop spec.
+	 */
+	constructor (props, spec) {
+		this.props = props;
 		this.spec = spec;
 
-		let name = this.spec.name;
+		// Register before any side effects, so re-entrant lookups during the
+		// init below (e.g. a dependent's getter reading `this.someOtherProp`)
+		// find this wrapper in the Map.
+		props.set(spec.name, this);
+
+		let { name, reflect, defaultProp, get: computed } = spec;
+		let { element } = props;
+
 		if (Object.hasOwn(element, name)) {
-			// A local data property will shadow the accessor that is defined on the prototype
+			// A local data property (e.g. a class field, or a pre-mount write
+			// that was shadowed onto the element) covers the prototype accessor.
 			// See https://github.com/nudeui/element/issues/14
 			let value = element[name];
-			delete element[name]; // Deleting the data property will uncover the accessor
-			element[name] = value; // Invoking the accessor means the value doesn't skip parsing
+			delete element[name];
+			this.set(value, { source: "property" });
 		}
-		// TODO this needs editing
-		else if (element.props[name] === undefined && !this.spec.defaultProp) {
-			// Is not set and its default is not another prop
-			this.changed(element, { source: "default", element });
+		else if (reflect.from && element.hasAttribute(reflect.from)) {
+			this.set(element.getAttribute(reflect.from), {
+				source: "attribute",
+				name: reflect.from,
+			});
 		}
+		else if (!defaultProp && !computed) {
+			// Plain prop (value-or-function default, no `get`, no `defaultProp`).
+			// Computed / defaultProp'd props don't need this — they receive their
+			// initial value via the cascade from their dependencies.
+			this.changed({ source: "default" });
+		}
+	}
+
+	get element () {
+		return this.props.element;
 	}
 
 	get name () {
 		return this.spec.name;
 	}
 
-	get type () {
-		return this.spec.type;
-	}
-
 	/**
-	 * Read this prop's current value for an element, falling back to its default.
-	 * @param {HTMLElement} element
+	 * Read this prop's current value, falling back to its default if unset.
 	 */
 	get () {
-		let element = this.element;
-		let value = this.value;
-
-		if (value === undefined) {
-			this.update(element);
-			value = this.value;
+		if (this.value === undefined) {
+			this.update();
 		}
 
-		if (value === undefined) {
-			if (this.defaultProp) {
-				value = this.defaultProp.get(element);
-			}
-			else if (this.default !== undefined) {
-				value = resolveValue(this.default, [element, element]);
-
-				try {
-					value = this.parse(value);
-				}
-				catch (e) {
-					console.warn(
-						"Failed to parse default value",
-						value,
-						`for prop ${this.name}. Original error was: `,
-						e,
-					);
-					return null;
-				}
-
-				if (this.spec.convert) {
-					value = this.spec.convert.call(element, value);
-				}
-			}
+		if (this.value !== undefined) {
+			return this.value;
 		}
 
-		return value;
+		let { spec, element } = this;
+		let raw;
+
+		if (spec.defaultProp) {
+			raw = this.props.forSpec(spec.defaultProp).get();
+		}
+		else if (spec.default !== undefined) {
+			raw = resolveValue(spec.default, [element, element]);
+		}
+		else {
+			return undefined;
+		}
+
+		let value;
+		try {
+			value = spec.parse(raw);
+		}
+		catch (e) {
+			console.warn(
+				"Failed to parse default value",
+				raw,
+				`for prop ${this.name}. Original error was: `,
+				e,
+			);
+			return null;
+		}
+
+		return this.convert(value);
 	}
 
 	/**
-	 * Write a value to an element: parse, convert, store, and reflect to attribute when applicable.
-	 * @param {HTMLElement} element
-	 * @param {*} value Raw value (string from an attribute, or any from a property write).
+	 * Write a value: parse, convert, store, and reflect to attribute when applicable.
+	 * @param {*} value Raw value (string from an attribute, any from a property write).
 	 * @param {{source?: string, name?: string, oldAttributeValue?: string | null}} [options]
 	 */
-	set (element, value, { source, name, oldAttributeValue } = {}) {
-		let oldValue = element.props[this.name];
-
-		let parsedValue;
+	set (value, { source, name, oldAttributeValue } = {}) {
+		let { spec, element } = this;
+		let parsed;
 
 		try {
-			parsedValue = this.parse(value);
+			parsed = spec.parse(value);
 		}
 		catch (e) {
 			// Abort mission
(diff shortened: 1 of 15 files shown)


src/plugins/props/util/ElementProp.js
```diff
@@ -0,0 +1,278 @@
...
+		this.changed(change);
+	}
+
+	/**
+	 * Apply the spec's `convert` hook, or pass the value through if none.
+	 * @param {*} value
+	 */
+	convert (value) {
+		let { convert } = this.spec;
+		return convert ? convert.call(this.element, value) : value;
```
DmitrySharabin, line 184: I wonder if we should also catch convert errors like we do when parsing the value. 🤔

src/plugins/props/util/ElementProp.js
```diff
@@ -0,0 +1,278 @@
...
+	 * Apply a change descriptor by writing through the matching attribute or property.
+	 * Used to mirror a change after the fact (e.g. to replay queued changes).
+	 * @param {Object} change
+	 */
+	applyChange (change) {
+		let { element, spec } = this;
+
+		if (change.source === "attribute") {
+			let attributeName = change.attributeName ?? spec.reflect.to;
+			let attributeValue = change.attributeValue ?? element.getAttribute(attributeName);
```
DmitrySharabin: If `change.attributeValue` is `null` or `undefined`, and the attribute has a value, it won't be removed, and we'll finish with a stale state. I thought we'd already fixed that. What am I missing?

src/plugins/props/util/ElementProp.js
```diff
@@ -0,0 +1,278 @@
...
+			this.applyChange({ ...change, source: "property" });
+		}
+	}
+
+	/**
+	 * Invoke the spec's `changed` hook and cascade to {@link ElementProps#propChanged}.
+	 * @param {Object} change
+	 */
+	changed (change) {
+		this.spec.changed?.call(this.element, change);
```
DmitrySharabin, line 227: Shouldn't we also catch errors with `changed` hooks? 🤔

src/plugins/props/util/ElementProp.js
DmitrySharabin: I know why it happened. We inlined the attribute reflection instead of routing it through `applyChange`. I'd suggest bringing it back. We'll keep the reflection logic in one place and fix the bug with stale attribute in one shot.

DmitrySharabin: This should be landed simultaneously with the change on L163.

```suggestion
			let attributeValue =
				change.attributeValue !== undefined
					? change.attributeValue
					: element.getAttribute(attributeName);
```

src/plugins/props/util/ElementProp.js
```diff
@@ -0,0 +1,278 @@
...
+				Object.assign(change, { attributeName, attributeValue, oldAttributeValue });
+
+				let ignored = this.props.ignoredAttributes;
+				ignored.add(attributeName);
+				if (attributeValue === null) {
+					element.removeAttribute(attributeName);
+				}
+				else {
+					element.setAttribute(attributeName, attributeValue);
+				}
```
DmitrySharabin:
```suggestion
				this.applyChange({ ...change, source: "attribute" });
```

src/plugins/props/util/Prop.js
```diff
@@ -58,350 +46,76 @@ let Self = class Prop {
...
-
-				try {
-					value = this.parse(value);
-				}
-				catch (e) {
-					console.warn(
-						"Failed to parse default value",
+					// Pre-mount write: install a shadowing data property. When
+					// ElementProps is later constructed, ElementProp#initialize
+					// will pick it up via shadow recovery (Object.hasOwn).
```
DmitrySharabin: Looks like a stale comment: `ElementProp` doesn't have `initialize`.

DmitrySharabin:
Spotted a bug by testing our main consumer—the color elements lib.

## Computed prop can read a sibling's raw, unparsed pre-mount value during mount

### Summary

When a consumer sets props **before connecting** an element (`createElement` → set props → `append`), a computed prop can observe a *sibling* prop's **raw, unparsed** value during the mount cascade — bypassing that prop's `parse`/`convert`.

### Mechanism

1. **Pre-mount writes become shadowing own-properties.** Before the element is connected, `element.props` doesn't exist yet, so the prop's setter [falls back to `Object.defineProperty(this, name, …)`](https://github.com/nudeui/element/blob/c85fb62261c4f12740ca47162620cdd1de1d4096/src/plugins/props/util/Prop.js#L94-L110), installing an **own data property** that shadows the prototype accessor. (This is the supported "set before connect" pattern.)

2. **Mount initializes props one at a time.** `ElementProps`'s constructor runs [`this.spec.forEach(spec => this.forSpec(spec))`](https://github.com/nudeui/element/blob/c85fb62261c4f12740ca47162620cdd1de1d4096/src/plugins/props/util/ElementProps.js#L57). Each `ElementProp` is allocated **and** initialized in the same step — its constructor [checks `Object.hasOwn(element, name)`, and only then reads the shadow, `delete`s it, and `set()`s it](https://github.com/nudeui/element/blob/c85fb62261c4f12740ca47162620cdd1de1d4096/src/plugins/props/util/ElementProp.js#L37-L44) (running `parse`/`convert`). So a prop's shadow is recovered only when *that prop's* `ElementProp` is constructed.

3. **The race.** Take a **computed** prop **C** that reads two props **A** and **B**, with **A** declared before **B**. **C** depends on both, but **A** initializes first, so [**A**'s change cascades to **C**](https://github.com/nudeui/element/blob/c85fb62261c4f12740ca47162620cdd1de1d4096/src/plugins/props/util/ElementProps.js#L122-L128) *before* **B**'s `ElementProp` has been constructed. **B**'s shadow data property is therefore still on the element, so `this.B` inside **C**'s getter returns the **raw write** — the read hits the own property (JS own-over-prototype precedence) and [never reaches the accessor](https://github.com/nudeui/element/blob/c85fb62261c4f12740ca47162620cdd1de1d4096/src/plugins/props/util/Prop.js#L83-L88) or `parse`. *(In the test below: **A** = `trigger`, **B** = `items`, **C** = `derived`.)*

> **Note:** the construction *order* is only the trigger — the root cause is the shadow itself. Even if every wrapper were allocated up front, `this.B` would still hit the own data property; own-over-prototype precedence means the accessor (the only path through `parse`) is never consulted until the shadow is `delete`d. So this isn't fixed by merely ensuring a dependent finds its wrapper in the Map mid-cascade.

### Symptom

The computed getter sees the wrong type/shape (e.g. a raw string where `parse` would have produced an array/object). In practice this often throws a `TypeError` when the getter calls a type-specific method on the still-raw value (e.g. `value.find(...)` on a string), aborting the rest of mount. Non-throwing cases may *self-heal* after mount (once **B** initializes, it re-cascades to **C**), so this is a **transient mount-time** violation.

### Invariant being violated

Once `element.props` is installed, a read of any prop must route through its accessor (and therefore `parse`/`convert`). A pre-mount shadow must never be observable as the prop's value after mount has begun.

### Regression test
We can add it to the installation tests (after reducing the description).

```js
{
	name: "Computed prop reading a sibling during mount sees its parsed value, not the raw pre-mount write",
	description:
		"A pre-mount property write installs a shadowing own-property that covers the " +
		"prototype accessor. `trigger` is declared before `items`, so when its mount " +
		"cascade runs `derived`'s getter, `items` is still uninitialized and its shadow " +
		"is intact. The sibling read must route through `items`'s parse (raw string -> " +
		"array), not return the raw shadow string. `observed` captures that first " +
		"cascade-time read; reordering the props would make the test pass without " +
		"reproducing the bug.",
	run () {
		let observed;

		let tag = defineElement({
			plugins: [propsPlugin],
			props: {
				trigger: { type: Number },
				items: {
					parse: value => (typeof value === "string" ? value.split(",") : value),
				},
				derived: {
					get () {
						if (this.trigger !== undefined) {
							observed ??= this.items;
						}

						return observed;
					},
				},
			},
		});

		let element = document.createElement(tag);
		element.trigger = 1;
		element.items = "x,y,z";
		document.body.append(element);

		element.remove();
		return observed;
	},
	expect: ["x", "y", "z"],
}
```

Current (buggy) output:

```
FAIL  Computed prop reading a sibling during mount … Got "x,y,z", expected ["x", "y", "z"]
```


**UPDATE:** Proposed fix (diff) from Claude

```diff
--- a/src/plugins/props/util/ElementProps.js
+++ b/src/plugins/props/util/ElementProps.js
@@ -20,6 +20,15 @@
 	ignoredAttributes = new Set();
 
 	/**
+	 * Pre-mount property writes (`element.foo = …`) captured at construction.
+	 * They live as shadowing OWN data properties on the element; we lift them
+	 * off before init so they can't intercept init-time reads (see constructor).
+	 * Consumed by each {@link ElementProp}'s init pass.
+	 * @type {Map<string, *>}
+	 */
+	shadowValues = new Map();
+
+	/**
 	 * Queue of propchange-style events awaiting dispatch while paused.
 	 * @type {PropChangeEvent[]}
 	 */
@@ -30,13 +39,15 @@
 	 * Pre-mount user writes (`element.foo = …`) and pre-mount attribute writes
 	 * (`setAttribute`) leave traces on the element itself — a shadowing data
 	 * property and an attribute, respectively — which each {@link ElementProp}'s
-	 * init pass picks up. So there's no separate `initialize()` step: the
-	 * constructor allocates every wrapper and then runs init on each.
+	 * init pass picks up. So there's no separate `initialize()` step.
 	 *
-	 * The split is internal (allocate-all → init-all) so cascades during init
-	 * always find existing wrappers in the Map — a wrapper materialized
-	 * mid-cascade would fire its own default event in addition to the
-	 * cascading update.
+	 * A shadowing data property covers the prototype accessor (normal JS
+	 * precedence), so reading the prop returns the raw, unparsed write instead
+	 * of routing through the accessor. If init is interleaved per prop, an
+	 * earlier prop's cascade into a computed prop can read a later prop whose
+	 * shadow hasn't been recovered yet — and get the raw value. So we lift every
+	 * shadow off the element up front (capturing it in {@link shadowValues}),
+	 * then materialize wrappers; init-time reads now always hit the accessor.
 	 *
 	 * @param {HTMLElement} element The element this collection belongs to.
 	 */
@@ -49,6 +60,16 @@
 		// accessor (which routes via `element.props`).
 		element.props = this;
 
+		// Lift pre-mount shadow data properties off the element before any init
+		// runs, so they can't intercept init-time reads of not-yet-materialized
+		// props. Each ElementProp recovers its value from `shadowValues`.
+		for (let [name] of this.spec) {
+			if (Object.hasOwn(element, name)) {
+				this.shadowValues.set(name, element[name]);
+				delete element[name];
+			}
+		}
+
 		// Materialize an ElementProp for every spec. Each constructor
 		// self-registers and runs its own init (shadow / attr / default-fire),
 		// cascading to dependents via {@link forSpec} if needed — so wrappers
--- a/src/plugins/props/util/ElementProp.js
+++ b/src/plugins/props/util/ElementProp.js
@@ -34,13 +34,12 @@
 		let { name, reflect, defaultProp, get: computed } = spec;
 		let { element } = props;
 
-		if (Object.hasOwn(element, name)) {
-			// A local data property (e.g. a class field, or a pre-mount write
-			// that was shadowed onto the element) covers the prototype accessor.
+		if (props.shadowValues.has(name)) {
+			// A local data property (e.g. a class field, or a pre-mount write)
+			// shadowed the prototype accessor. ElementProps lifted it off the
+			// element before init and stashed the raw value here.
 			// See https://github.com/nudeui/element/issues/14
-			let value = element[name];
-			delete element[name];
-			this.set(value, { source: "property" });
+			this.set(props.shadowValues.get(name), { source: "property" });
 		}
 		else if (reflect.from && element.hasAttribute(reflect.from)) {
 			this.set(element.getAttribute(reflect.from), {
```

LeaVerou:
The fix seems much simpler to me:
1. Get these props through convert when you get a chance to. Even explicit prop writes should get through that *anyway*
2. Make sure a computed's dependencies are constructed before going any further

src/plugins/props/util/ElementProp.js
LeaVerou: Yes, but If we didn't before I'd say it's out of scope for this PR

src/plugins/props/util/ElementProp.js
LeaVerou: Same comment as above

PR #126, opened 2026-05-21
Replace happy-dom with browser-shims

DmitrySharabin:
## Summary

- Replace happy-dom with [`browser-shims`](https://github.com/nudeui/browser-shims) — per-class DOM stubs that extend native `EventTarget`, no `Event`/`CustomEvent` replacement, zero hTest conflicts
- Loaded via hTest's `setup` config (Node-only, guarded by `loadIf`)

### Dependency changes

| Package | Change |
|---|---|
| `@happy-dom/global-registrator` | Removed |
| `parse5` | Removed (transitive dep of browser-shims) |
| `htest.dev` | `^0.0.24` → `^0.0.25` (adds `setup` config) |
| `browser-shims` | Added (devDep) |

## Test plan

- [x] `npm test` → 103/107 pass, 4 skipped (matches baseline)
- [x] Standalone file: `npx htest test/plugins/props/inheritance.js` passes independently
- [x] No `happy-dom` or `restoreNativeCustomEvent` references remain in test/


LeaVerou (changes requested):
Can you move the DOM polyfill here and publish it as a separate package? https://github.com/nudeui/browser-shims

I'd split the different classes into separate files. Then, each file should check if the class is already available on `globalThis` and add it if not, or export that built-in class if so.

DmitrySharabin:
> Can you move the DOM polyfill here and publish it as a separate package? https://github.com/nudeui/browser-shims
> 
> I'd split the different classes into separate files. Then, each file should check if the class is already available on `globalThis` and add it if not, or export that built-in class if so.

Done. Now, the PR is tiny.

LeaVerou (approved):

PR #127, opened 2026-05-22
PropType class

LeaVerou:
## Summary

Rewrite of the prop-type system. Types are now first-class `PropType` instances composed via JS prototype-chain inheritance, replacing the previous design where types were entries in an opaque map of `{equals, parse, stringify}` method packs keyed on JS constructor.

The motivating problem: with method-pack registrations, there was no Type *instance*, no inheritance, no shared behavior between Array/Set or Object/Map without per-method duplication. Derivative configurations like `{is: Array, values: Number}` were re-resolved through ad-hoc spec merging on every lookup.

## Architecture

**One class.** `PropType` is the sole class. Type *instances* hold their spec verbatim and use `Object.create(parent)` for derivatives — every option lookup walks the JS prototype chain, no merging, no copying.

**Registry, single map.** Keyed on JS constructor (`is`) for concrete types and on string `name` for abstract types. `PropType.for(input, {fallback})` resolves any user-facing identifier (PropType instance, constructor, string, spec object) to the right instance: registered singleton, fresh derivative for specs with extras, or fallback.

**Abstract vs concrete.** Abstracts (currently just `IterableType`) are registered with `name` and no `is` — they exist to be inherited from, not to claim a JS constructor. Concrete types use `extends: <abstract>` to decouple their chain parent from `is`, e.g. `{is: Array, extends: IterableType}`.

**One dispatch chain.** `PropType.prototype.equals/parse/stringify` walk `obj.super`, dispatching to the first override found on `obj.spec`. The same walker handles auto-wrapped helper methods (e.g. `parseItems`, `parseEntries`), so abstracts can publish arbitrary helpers and descendants invoke them as plain `this.x(...)`.

**Streaming pipeline.** `IterableType.parseItems` (raw splitter) and `IterableType.parse` (typed) are both generators. `MapType` extends `IterableType`, reusing inherited `this.parseItems(value)` for raw string parts and splitting each on `:` in its own `parseEntries` to yield `[key, value]` tuples. Concrete types (`array.js`, `set.js`, `map.js`, `object.js`) consume the iterator with the appropriate container constructor (`Array.from`, `new Set`, `new Map`, `Object.fromEntries`). Each input value flows through the chain exactly once, no intermediate arrays.

**Map IS the dictionary.** There's no separate `DictionaryType` abstract — `Map` is the canonical JS dictionary, `Object` a constrained realization. `ObjectType extends MapType`, inheriting the entry-parsing pipeline.

## Spec keys

| Key            | Role                                                                                                         |
| -------------- | ------------------------------------------------------------------------------------------------------------ |
| `is`           | JS constructor this type produces; doubles as registry key. Optional for abstracts.                          |
| `extends`      | Explicit chain parent. Lets the parent differ from `registry.get(is)` (e.g. concretes extending an abstract). Accepts a `PropType` instance or a name string. |
| `name`         | Registry key for abstracts that have no `is`.                                                                |
| `subTypes`     | Spec keys whose values are themselves type specs (`["values"]` for Iterable, `["keys", "values"]` for Map). Resolved to `PropType` instances at construction; unspecified keys default to `PropType.any` (inherited via the prototype chain), so `this.values.parse(v)` works unconditionally. |
| `equals`/`parse`/`stringify` | Method overrides with shared null/identity short-circuits.                                       |
| custom methods | Auto-wrapped at construction into super-walking dispatchers, callable as `this.x(...)`.                      |

## Public API additions

- `PropType.for(input, { fallback })` — universal type resolver.
- `PropType.register(spec)` — register a built-in or custom type.
- `instance.isA(otherType)` — chain membership check (replaces the `instanceof` checks that don't apply when abstracts aren't JS classes).
- `import { PropType, Iterable } from "nude-element/props"` for extension. Other built-in singletons can be reached via `PropType.for(name)`.
- `Symbol.toStringTag` on `PropType` so `console.log(arrayType)` shows `ArrayType`.

## File layout (under `src/plugins/props/`)

- `util/PropType.js` — the class.
- `util/split.js` — shared string splitter (pair-aware, fault-tolerant).
- `types/iterable.js` — abstract.
- `types/array.js`, `set.js`, `map.js`, `object.js` — one file per concrete iterable/dictionary type.
- `types/basic.js` — `Boolean`, `Number`, `Function`.

## Behavior changes worth flagging

- The auto-spacing heuristic in `joinItems` is gone — custom separators are used verbatim. Default separator stays `", "` so the common case is unchanged.

## Test plan

- [x] `npm test` — 163/167 pass (4 pre-existing skips, 0 failures).
- [x] New `test/PropType.js` covers `for()` / `register()` / derivation / dispatch / built-in primitives / list & dict behavior / nested composition / `isA`.
- [x] Manual sanity check on a consumer using `extends: IterableType`.
- [x] Confirm Map/Object derivatives with custom `keys`/`values` still round-trip cleanly.


src/plugins/props/util/PropType.js
```diff
@@ -0,0 +1,333 @@
...
+		}
+	}
+
+	/**
+	 * @param {unknown} a
+	 * @param {unknown} b
+	 * @returns {boolean}
+	 */
+	equals (a, b) {
+		if (a === null || b === null || a === undefined || b === undefined) {
```
DmitrySharabin: Nit: What if we make it a bit shorter via coercion?
```suggestion
		if (a == undefined || b == undefined) {
```

src/plugins/props/util/PropType.js
```diff
@@ -0,0 +1,333 @@
...
+
+		return typeof a.equals === "function" ? a.equals(b) : false;
+	}
+
+	/**
+	 * @param {unknown} value
+	 * @returns {unknown}
+	 */
+	parse (value) {
+		if (value === null || value === undefined) {
```
DmitrySharabin: And here:
```suggestion
		if (value == undefined) {
```

src/plugins/props/util/PropType.js
```diff
@@ -0,0 +1,333 @@
...
+		return callableBuiltins.has(Type) ? Type(value) : new Type(value);
+	}
+
+	/**
+	 * Null/undefined produce `null` (signaling attribute removal).
+	 * @param {unknown} value
+	 * @returns {string | null}
+	 */
+	stringify (value) {
+		if (value === null || value === undefined) {
```
DmitrySharabin: And here:
```suggestion
		if (value == undefined) {
```

src/plugins/props/util/PropType.js
```diff
@@ -0,0 +1,333 @@
...
+
+	/**
+	 * Is this type a kind of `other` — i.e. is `other` somewhere in the
+	 * super chain (or is it `this` itself)? Replaces `instanceof` checks
+	 * that no longer apply now that abstract types are PropType instances
+	 * rather than JS classes.
+	 * @param {PropType} other
+	 * @returns {boolean}
+	 */
+	isA (other) {
```
DmitrySharabin, line 201: I'm a bit confused by this name. For me, it's read like “is 'A'” (not as an indefinite article; probably, it's just me).
How about `kindOf` (as it is referenced in the JSDoc) or `instanceOf` (to signal the similar to `instanceof` semantics)? Or `isInstanceOf` (inspired by `isPrototypeOf`), but it's a bit too wordy, to my taste.

src/plugins/props/util/PropType.js
```diff
@@ -0,0 +1,333 @@
...
+	 * anything else (including named-only abstracts like `"Iterable"`) passes
+	 * through as the bare string.
+	 * @param {Function | string | undefined} is
+	 * @returns {Function | string | undefined}
+	 */
+	static normalizeIs (is) {
+		if (typeof is !== "string") {
+			return is;
+		}
+		let resolved = globalThis[is];
```
DmitrySharabin, line 302: I wonder if `globalThis` is defined in every environment the library might be used. Can't we stumble on a case where this code might throw? 🤔

src/plugins/props/util/split.js
```diff
@@ -0,0 +1,103 @@
+export const defaultPairs = {
+	nest: {
+		"(": ")",
+		"[": "]",
+		"{": "}",
+	},
+	ignore: {
+		'"': '"',
+		// "'": "'",
+		// "`": "`",
```
DmitrySharabin, line 10: I wonder if it's the right time to unblock (add support) for this as well. 🤔

src/plugins/props/README.md
```diff
@@ -153,6 +153,62 @@ While `defaultKey` _can_ be a non-function, this is almost never what you want,
...
 It can be either a constant (e.g. `true`) or a function, in which case it’s passed the key and the index as arguments.
 
+#### Custom types
+
+Types are *instances* of the single `PropType` class. Each instance carries the spec it was constructed with — its constructor (`is`), any type options, and any `equals` / `parse` / `stringify` overrides — and is shared across every prop that references it. An abstract type (`Iterable`) is a `PropType` instance registered by `name` rather than `is`; concrete types declare `extends: <abstract>` to inherit its parsing behavior via the JS prototype chain (e.g. `Array`, `Set`, and `Map` all extend `Iterable`, and `Object` extends `Map`).
+
+For most custom types, the simplest path is to register a spec object directly. Methods read options from `this.spec`:
+
+```js
+import { PropType } from "nude-element/plugins";
```
DmitrySharabin: `PropType` is not reachable from this path (but it is reachable from `nude-element`).

I wonder if it's correct to import it from `plugins`. It doesn't sound like a plugin to me. But it's completely fine that it's a part of the lib itself, and the import from `nude-element` looks fine for me.

Should we fix the import/export in the source code or fix the docs (also on L176, L190)?

src/plugins/props/types/README.md
```diff
@@ -0,0 +1,81 @@
...
+Both are generators. Concrete types consume them with the appropriate terminal container constructor (`Array.from`, `new Set`, `new Map`, `Object.fromEntries`) so each input value flows through the chain exactly once — no intermediate arrays.
+
+To call a parent's method from inside an override, use `ParentType.spec.method.call(this, …args)`. Doing it via `this.super.method(…)` would re-bind `this` to the parent and lose access to your derivative's `this.values` / `this.spec.separator`.
+
+## A parametrized custom type
+
+For something more involved than the simple [`Color` example](../README.md#custom-types), here's a `Length` type that accepts a `unit` option, demonstrating how a single registration becomes the basis for many derivatives:
+
+```js
+import { PropType } from "nude-element/plugins";
```
DmitrySharabin: If the answer to [this question](https://github.com/nudeui/element/pull/127/changes#r3300048235) is that we should fix the docs, then this line should also be fixed.

DmitrySharabin:
I left a few comments and proposals and submitted a [PR](https://github.com/nudeui/element/pull/128) with a few more fixes.

I also tested it with the color elements. There was only one change needed to adopt the changes introduced by this PR: to register the `Color` type before defining the `ColorElement` class.

src/plugins/props/types/util.js
DmitrySharabin, line 1: This file is dead code now. Should we just remove it?

DmitrySharabin:
A few spots where the docs describe the mechanism slightly differently from what the code does. The user-facing *behavior* described is correct in each case — it's the "how" that's off.

---

### 1. "Auto-wrapped … super-walking dispatcher" (spec-keys table)

[`types/README.md` L30](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/types/README.md#L30):

> any other method — Auto-wrapped at construction into a super-walking dispatcher, callable as `this.x(…)` from anywhere in the chain.

In `init()` ([`PropType.js` L126-132](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/util/PropType.js#L126-L132)), non-prototype keys are assigned directly (`this[key] = spec[key]`) — no wrapping, no dispatcher. They're inherited by derivatives through the plain JS prototype chain (`Object.create(parent)`), which is why `this.x(…)` works from anywhere in the chain. But the mechanism is prototype inheritance, not auto-wrapping.

Possible fix: "Stored as own properties at construction and inherited by derivatives via the prototype chain, so `this.x(…)` resolves to the nearest ancestor's implementation."

---

### 2. Architecture paragraph — "The dispatcher walks `obj.super`"

[`types/README.md` L81](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/types/README.md#L81):

> The dispatcher walks `obj.super` for each method (`equals` / `parse` / `stringify` plus any custom helpers), invoking the first `spec[method]` it finds with `this` bound to the original caller.

`obj.super` is only walked by `isA()`. The `equals`/`parse`/`stringify` prototype methods read `this.spec_<name>`, which resolves through the JS prototype chain set up by `Object.create` — finding the first override in the chain. Custom helpers are stored as own properties on the registered instance and inherited the same way.

Possible fix: "Each prototype method (`equals`/`parse`/`stringify`) reads `this.spec_<name>`, which resolves through the JS prototype chain set up by `Object.create(parent)` — the first override in the chain wins. Custom helpers (`parseItems`, `parseEntries`) are own properties that propagate to derivatives the same way, so `this.x(…)` always finds the nearest ancestor's implementation."

---

### 3. MapType in the "Abstract helpers" table

[`types/README.md` L37](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/types/README.md#L37) and the section heading "Abstract helpers":

MapType is a concrete type (`is: Map` — [`map.js` L17-18](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/types/map.js#L17-L18)), not an abstract. It does serve as a base for Object, but it has an `is`. The table heading "Abstract helpers" groups it with the genuinely abstract `Iterable`.

Also, `MapType` is a local variable in `map.js` — it's not exported under that name from any public entry point. `nude-element/plugins/types` exports it as `Map`. The same applies to [`props/README.md` L173](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/README.md#L173), which recommends `extends: MapType` in prose.

Possible fix: rename the section to something like "Helpers from base types", and either export `MapType` as a named export from `types/index.js`, or refer to it as `Types.Map` in the docs (matching what users can actually import).

---

### 4. Terminal constructors — "`Array.from`"

[`types/README.md` L39](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/types/README.md#L39):

> Concrete types consume them with the appropriate terminal container constructor (`Array.from`, `new Set`, `new Map`, `Object.fromEntries`)

Array uses spread (`[...this.parsedItems(value)]` — [`array.js` L8](https://github.com/nudeui/element/blob/f0cd471eb68dcd9bfbb9734dca28f7a8a6a3bf38/src/plugins/props/types/array.js#L8)), not `Array.from`. Set/Map use `new this.is(…)`.

Possible fix: "an appropriate terminal consumer (spread, `new this.is(…)`, `Object.fromEntries`)" or just drop the parenthetical list.

---

_— flagged by @claude_

src/plugins/props/util/PropType.js
LeaVerou: Claude tried to do that, I'd rather avoid `==` as it's too unpredictable.

src/plugins/props/util/PropType.js
LeaVerou: "is a" is a common way to reference this type of relationship, but I'm not opposed to something else. But `kindOf` feels more vague and to me `instanceof` implies constructor, not inheritance.

src/plugins/props/util/PropType.js
LeaVerou: I think it's fine

src/plugins/props/util/split.js
LeaVerou: It's not about adding support, it's about how likely these pairs are to be used in a way that isn't a pair. E.g. single quotes are often used instead of apostrophes. I think it's fine leaving this to consumers and make it their responsibility when it fails, especially since now they can provide `pairs` alongside a custom type.

src/plugins/props/README.md
LeaVerou: It's part of Props, which is a plugin. But perhaps `nude-element/props` would make more sense. And eventually moving props to a separate package even…

LeaVerou: But this is definitely wrong. The only named exports of `nude-element/plugins` would be the plugins, not parts of plugins.

LeaVerou:
> I left a few comments and proposals and submitted a [PR](https://github.com/nudeui/element/pull/128) with a few more fixes.
> 
> I also tested it with the color elements. There was only one change needed to adopt the changes introduced by this PR: to register the `Color` type before defining the `ColorElement` class.

Could you elaborate on this? It should not need registration, the `any` type should work fine for that.

DmitrySharabin:
> > I left a few comments and proposals and submitted a [PR](https://github.com/nudeui/element/pull/128) with a few more fixes.
> > I also tested it with the color elements. There was only one change needed to adopt the changes introduced by this PR: to register the `Color` type before defining the `ColorElement` class.
> 
> Could you elaborate on this? It should not need registration, the `any` type should work fine for that.

The issue is with `parse`. In the old system, [`generic.parse(value, type)`](https://github.com/nudeui/element/blob/ba1d2107d735ef89e0f9cf7df9e6eba5dc0c5e48/src/plugins/props/types/basic.js#L28-L34) received the type spec as a second argument, so even for unregistered constructors it could do `new Type(value)`. In the new system, [`PropType.for(Color)`](https://github.com/nudeui/element/blob/7ee136e4763a553d04fba6c5bb56aab79d0b1c4c/src/plugins/props/util/PropType.js#L280-L293) looks up `Color` in the registry, doesn't find it, and returns `PropType.any` — which has no `is`, so [`parse()` just passes the value through as-is](https://github.com/nudeui/element/blob/7ee136e4763a553d04fba6c5bb56aab79d0b1c4c/src/plugins/props/util/PropType.js#L183-L185).

Several color-elements props declare `type: Color`, relying on auto-instantiation from attribute strings. Without registration, those props hold the raw string `"red"` instead of a `Color` instance.

`PropType.register({ is: Color })` restores the old behavior by giving the type an `is` that `parse()` can use for [`new Color(value)`](https://github.com/nudeui/element/blob/7ee136e4763a553d04fba6c5bb56aab79d0b1c4c/src/plugins/props/util/PropType.js#L188).

That said, this could also be fixed upstream — `PropType.for()` could auto-create a minimal `PropType({ is: Constructor })` for unregistered constructors instead of falling back to `any`, which would match the old behavior without requiring consumers to register every custom class.

This bug is the last blocker.

src/plugins/props/util/PropType.js
DmitrySharabin: Let's leave it as-is, then. Especially if it's a common thing.

DmitrySharabin:
The issue with unregistered constructors is fixed. Verified with the color elements: No need to register `Color` anymore.

DmitrySharabin:
## Docs-vs-code review

Flagged by Claude and Codex.

### P2 — `this.super` warning is backwards

[`types/README.md` line 41](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/README.md#L41) says:

> Doing it via `this.super.method(…)` would re-bind `this` to the parent and lose access to your derivative's `this.values` / `this.spec.separator`.

The super-proxy does the opposite — it [explicitly preserves the child's `this`](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/util/PropType.js#L340):

```js
let thisArg = this === receiver ? self : this;
return val.apply(thisArg, args);
```

The tests [confirm this works correctly](https://github.com/nudeui/element/blob/84c97db/test/PropType.js#L282-L299). The recommended workaround (`ParentType.spec.method.call(this, …args)`) hard-codes the parent and bypasses the `get_` wrappers (e.g. null-handling in `get_parse`), so it's the worse option. This will actively mislead users — suggest flipping the guidance to recommend `this.super.method(…)`.

### P2 — Stale `spec_` prefix in PropType JSDoc

[`PropType.js` lines 28–41, 48](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/util/PropType.js#L28-L41) describe a `spec_` storage/dispatch mechanism:

> "overrides that collide with prototype names … are stored under a `spec_` prefix" / "then call `this.spec_<name>(…)` if present"

No `spec_` properties exist in the code. The actual mechanism is `get_` transforms (`get_equals`, `get_parse`, `get_stringify`) that close over the spec function in a wrapper. If these JSDoc comments feed generated API docs, they'll describe an architecture that doesn't exist.

### P2 — Non-existent import path in PR description

The PR body claims:

> `import * as Types from "nude-element/plugins/types"` exports the built-in singletons under their JS constructor names

There's no `"./plugins/types"` in [`package.json` exports](https://github.com/nudeui/element/blob/84c97db/package.json#L24-L29). The actual export is `"./props"` → `src/plugins/props/index.js`, and [`types/index.js`](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/index.js) only exports `PropType` and `Iterable`, not named singletons like `Types.Array`.

### P3 — Props README type-options table is stale

The [existing table at `README.md` line 136](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/README.md#L136-L143):

- Lists `keys` as applying only to `Map`, but `Object` now [extends `MapType`](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/object.js#L4) and inherits `subTypes: ["keys", "values"]`.
- Omits `separator`, `joiner`, and `pairs`, which are used by all iterable/dict types.
- The types/README [built-in types table](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/README.md#L7-L15) has the correct info. Either update this table to match or make it a brief summary that explicitly defers to the full reference.

### P3 — `Array.from` in docs vs spread in code

Both the [types/README line 39](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/README.md#L39) and the PR description claim `Array.from` as a terminal constructor, but [`array.js`](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/array.js#L8) uses `[...this.parsedItems(value)]`.

### P3 — Length example teaches `this.spec.unit` instead of `this.unit`

The [example at `types/README.md` line 53](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/types/README.md#L53) uses `this.spec.unit ?? "px"`. Since `init()` [lifts all spec keys onto the instance](https://github.com/nudeui/element/blob/84c97db/src/plugins/props/util/PropType.js#L123), the idiomatic access is `this.unit` — consistent with how all built-in types access their options (`this.separator`, `this.defaultKey`, etc.). `this.spec.unit` works but will confuse anyone reading the built-in source.

DmitrySharabin (approved):
I think we fixed all that we spotted during the reviews. If not, we'll fix it in the follow-up work. LGTM!

PR #128, opened 2026-05-25
Handle escaped separators in split()

DmitrySharabin:
## Summary

`split()` honors `\<sep>` as an escaped separator: not split on, and the backslash is stripped from the yielded part. `MapType.parseEntries()` uses it for `:`-splitting.

## Test plan

- [x] `npx htest test/split.js --ci`
- [x] `npx htest test/PropType.js --ci`

### DIFF
--- a/src/plugins/props/types/map.js
+++ b/src/plugins/props/types/map.js
@@ -1,7 +1,6 @@
 import PropType from "../util/PropType.js";
+import { split } from "../util/split.js";
 import Iterable from "./iterable.js";
-
-const entrySplitter = /(?<!\\):/;
 
 /**
  * Concrete type for `Map`, which also serves as the canonical dictionary in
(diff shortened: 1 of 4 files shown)


src/plugins/props/types/basic.js
```diff
@@ -1,6 +1,6 @@
 import PropType from "../util/PropType.js";
 
-PropType.register({
+export const BooleanType = PropType.register({
```
LeaVerou: What is the use case for importing these separately? For extending, we can do `extends: "Boolean"`. To access them, `PropType.for("Boolean")` is easy.

LeaVerou (changes requested):
I think since escaped colons are an older bug, they should remain separate. For the exports, I'm not sure this is the right fix. Can you commit the JSdoc fixes directly to types-refactor and slim this PR down to the colon handling fix?

LeaVerou:
> Can you commit the JSdoc fixes directly to types-refactor

Done

DmitrySharabin:
> slim this PR down to the colon handling fix

Done. After the type refactors PR is landed, we'll retarget it to the `rollback-signals` branch (if it won't happen automatically).

LeaVerou (changes requested):
I think instead of handling this ad hoc, `parseEntries()` should be using `split()` which I believe handles escapes already (and if it doesn't, it should).

Issue #125, opened 2026-05-21
Computed props with defaultProp dependency return stale cached values


DmitrySharabin:
## Problem

`ElementProp.update(dependency)` short-circuits when the triggering dependency is the prop's `defaultProp`:

```js
if (dependency && dependency.spec === spec.defaultProp) {
    this.changed({ source: "default" });
    return;  // ← getter never re-runs
}
```

This works for pure passthrough props, but breaks when a prop declares **both** `defaultProp` and `get` for the same dependency. The getter's cached `this.value` goes stale after the first evaluation.

### Reproduction

A prop like:

```js
space: {
    get () { return this._el.space_picker.selectedSpace; },
    defaultProp: "spaceId",
    dependencies: ["spaceId"],
}
```

1. First read → `value` is `undefined` → `get()` runs → caches result (e.g. `oklch`)
2. `spaceId` changes to `"hsl"` → cascade calls `space.update(spaceIdProp)`
3. `defaultProp` early-return fires → `changed({ source: "default" })` notifies listeners but **does not invalidate `this.value`**
4. Next read of `space` → `this.value` is still `oklch` → stale value returned

### Current workaround

The consumer manually invalidates the cache:

```js
let spaceProp = this.props.get("space");
if (spaceProp.value && spaceProp.value.id !== value) {
    spaceProp.value = undefined;
}
```

This works but depends on internal field access.

### Proposed fix

When the defaultProp early-return fires and the prop has a `get`, invalidate the cached value so the getter re-evaluates lazily on next read:

```js
if (dependency && dependency.spec === spec.defaultProp) {
    if (spec.get) {
        this.value = undefined;
    }
    this.changed({ source: "default" });
    return;
}
```

Eager re-evaluation (calling the getter immediately) doesn't work here because the `changed` callback may need to update DOM state that the getter reads from.

Tested against `color-elements`'s `<color-picker>` — the fix eliminates the consumer-side cache invalidation and all space-change paths work correctly.

LeaVerou:
If the fix is that small that's fine, but in general I suspect these types of issues will go away with the changes we're planning.

Please make sure to add a test for this!

DmitrySharabin:
> Please make sure to add a test for this!

I have this test locally. I'll add it as soon as we have all the pending changes merged. I'll copy it here just in case.

```js
import { default as propsPlugin } from "../../../src/plugins/props/index.js";
import { defineElement } from "../../util/dom.js";

export default {
[...]

PR #129, opened 2026-05-27
Add batched `propschange` event

LeaVerou:
## Summary

Adds a new `propschange` event: a coalesced, microtask-deferred batch event that fires once after a burst of `propchange` events settles, carrying the net first→last delta per prop as `Map<name, oldValue>`. Round-trips (value returns to its burst-start) drop out. Lives in a new `propschange` sub-plugin under `props/` and is auto-wired to subclasses that define an `updated(event)` method, mirroring Lit's `updated(changedProperties)`.

Mount fires one `propschange` with every prop in `changed` (`oldValue: undefined`), so initial values arrive as a single settled snapshot.

### New: `propschange` event + plugin (`src/plugins/props/propschange.js`)

Listens for `propchange` on each instance, maintains per-prop burst tracking (`firstOldValue` + latest `value` + spec), and fires one `PropsChangeEvent("propschange", { changed })` per microtask. Auto-wires `updated(event)`. The `props/` plugin folder now mirrors the `events/` pattern: `base.js` is the core props plugin, `index.js` is a meta-plugin bundling `base + propschange`, so the default `NudeElement` ships with both. Anyone who wants props core in isolation can import `props/base.js`.

### New: `ElementProp.source`

Tracks where the current value came from: `"default" | "property" | "attribute" | "get" | "convert"`. Drives the "user-owned" stickiness ElementProps uses to keep a `"get"` recompute from silently stripping ownership that a prior property/attribute write established.

### Breaking: `PropChangeEvent` shape

`PropChangeEvent` now extends `Event` (not `CustomEvent`) and exposes its fields directly — `e.value`, `e.oldValue`, `e.source`, `e.attributeName`, etc. — instead of nesting under `e.detail`. Each dispatch gets a fresh event object; fields don't mutate past the handler.

Downstream impact (incl. color-elements): `e.detail.*` → `e.*` on `propchange` listeners.

### Behavior: paused/resume sequential coalescing

`paused`/resume themselves already existed on main, but on `paused = false` the queue used to flush every queued event verbatim. Now it walks the queue in write order and **in-place coalesces consecutive same-prop entries** (preserving the run's burst-start `oldValue`), dispatching one event per run. A sequence like `A A B A` produces three dispatches, not two — non-adjacent same-prop runs stay separate, so the consumer hears a faithful narrative.

### Behavior: `on*=` handlers in markup

On*-prop handlers now attach via the on* prop's spec `changed` callback, which runs synchronously inside `ElementProp.changed` — before `propChanged` fires. Handlers bind during the prop's own init pass, in time for the mount propchanges that follow. Markup `<x onfoo="…">` receives the mount-time event with its real `source` (e.g. `"default"`) rather than a synthesized `"initial"` re-fire.

## Tests
- [x] `npx htest test/index.js` — 185/189 pass (4 pre-existing skips)
- [ ] Update color-elements for the `e.detail` → `e.*` flat-shape change and sanity-check the `propschange` / `source`-state behavior


src/plugins/props/util/ElementProp.js
```diff
@@ -206,9 +252,12 @@ export default class ElementProp {
...
 		if (dependency && dependency.spec === spec.defaultProp) {
-			// The default prop changed; the resolved default may differ, but since
-			// we have no value of our own to update, just notify listeners.
-			this.changed({ source: "default" });
+			let oldValue = this.value;
+			this.value = dependency.value === undefined
+				? undefined
+				: this.convert(spec.parse(dependency.value));
+			this.source = "default";
+			this.changed({ source: "default", value: this.value, oldValue });
```
DmitrySharabin: Nit: Is there a reason why not `source: this.source`? If we rename the source one day, we'll have to do two changes instead of one.

DmitrySharabin:
> ### Breaking: `PropChangeEvent` shape
> 
> `PropChangeEvent` now extends `Event` (not `CustomEvent`) and exposes its fields directly — `e.value`, `e.oldValue`, `e.source`, `e.attributeName`, etc. — instead of nesting under `e.detail`. Each dispatch gets a fresh event object; fields don't mutate past the handler.
> 
> Downstream impact (incl. color-elements): `e.detail.*` → `e.*` on `propchange` listeners.

I wonder if we should mention it in the docs just to draw the user's attention to the fact that we don't use `event.details` like they might expect for custom events. 🤔

src/plugins/props/util/ElementProps.js
```diff
@@ -164,18 +159,52 @@ export default class ElementProps extends Map {
...
+	#flushPending () {
+		// In-place coalesce consecutive same-prop entries: collapse later
+		// events into the run's first one, keeping its burst-start oldValue.
+		let writeIdx = 0;
+		let lastName = null;
+		for (let event of this.#eventQueue) {
+			if (event.name === lastName) {
+				let prev = this.#eventQueue[writeIdx - 1];
+				let { oldValue } = prev;
+				Object.assign(prev, event);
```
DmitrySharabin: If, during the pause, we set a prop value via a reflected attribute and then via a property, the dispatched event (with `source === "property"`) will have wrong (stale) properties from the previous event (with `source === "attribute"`), i.e., `attributeName`, `attributeValue`, and `oldAttributeValue`.

To reproduce:

```js
element.props.paused = true;
element.setAttribute("foo", "1");
element.foo = 2;
element.props.paused = false;
```

DmitrySharabin: Another issue with this code is that `PropChangeEvent` extends `Event`, and `Event.isTrusted` is a read-only getter on `Event.prototype`. `Object.assign` throws on every event. We probably shouldn't be that liberal with in-place coalescing.

src/plugins/props/util/ElementProp.js
```diff
@@ -206,9 +252,12 @@ export default class ElementProp {
 		let { spec } = this;
 
 		if (dependency && dependency.spec === spec.defaultProp) {
```
DmitrySharabin: Fixes issue #125. I think it's worth doing it here. The regression test is on `main` (and can be here after the rebase).

```suggestion
		if (dependency && dependency.spec === spec.defaultProp && !spec.get) {
```

DmitrySharabin (changes requested):
We have a couple of issues to fix. See comments.

src/plugins/props/util/ElementProp.js
LeaVerou: No, feel free to do it!

src/plugins/props/util/ElementProps.js
LeaVerou: Yikes, wtf Claude?! Pushing a fix now.

DmitrySharabin:
I'd also suggest adding these two regression tests:

```js
{
      name: "Round-trip on one prop drops it; genuinely changed prop stays",
      async run () {
              let { element } = this.data;
              await flush(); // drain mount-time propschange
              this.data.propsEvents.length = 0;

              element.a = 5;
              element.a = 0; // back to stored default
              element.b = "new";

              await flush();
              return this.data.propsEvents;
      },
      arg: {
              props: {
                      a: { type: Number, default: 0 },
                      b: { type: String, default: "old" },
              },
      },
      expect: [[["b", "old"]]],
},
{
      name: "Re-entrant write from a propchange handler joins the same propschange",
      async run () {
              let { element } = this.data;
              await flush(); // drain mount-time propschange
              this.data.propsEvents.length = 0;

              element.addEventListener("propchange", e => {
                      if (e.name === "a") {
                              element.b = "from_listener";
                      }
              });

              element.a = "new";
              await flush();
              return this.data.propsEvents;
      },
      arg: { props: { a: { default: "foo" }, b: { default: "bar" } } },
      expect: [[["a", "foo"], ["b", "bar"]]],
}
```

DmitrySharabin (approved):
I'd suggest adding [these tests](https://github.com/nudeui/element/pull/129#issuecomment-4565987850) (I believe they might help catch regressions). Other than that, LGTM! Excited to see what comes next.

src/plugins/props/util/ElementProp.js
```diff
@@ -70,53 +97,77 @@ export default class ElementProp {
...
 	 * @param {*} value Raw value (string from an attribute, any from a property write).
 	 * @param {{source?: string, name?: string, oldAttributeValue?: string | null}} [options]
 	 */
 	set (value, { source, name, oldAttributeValue } = {}) {
+		if (source !== "property" && source !== "attribute") {
+			// Non-user sources don't write through internalValue; they're
+			// derivations. Route to the cascade entry point.
+			this.update();
+			return;
+		}
```
DmitrySharabin: As I can see from the code, any source that is not related to user-supplied values will be discarded (I couldn't find a place where this branch leads to `this.source = source`). Is it intentional? Does it mean that custom sources can't be introduced? If so, L46-47 are misleading (or I'm reading them wrong).

src/plugins/props/util/ElementProp.js
```diff
@@ -65,54 +97,78 @@ export default class ElementProp {
...
+			//
+			// Edge case: a computed prop with no declared deps never receives
+			// a cascade, so a bootstrap seed stays as the visible value forever
+			// — spec.get is effectively masked. Pre-mount writes to depless
+			// computeds are pathological config; the alternative (silently
+			// dropping the seed) seems worse.
+			let raw =
+				this.internalValue !== undefined
+					? this.internalValue
+					: spec.parse(spec.get.call(element));
```
DmitrySharabin: Just for me: we don't wrap `parse` with `try/catch` here (and on L154) to fail early on incorrect prop specs? To let the author know asap that they do something wrong?

src/plugins/props/util/ElementProp.js
LeaVerou: Pushed a fix

src/plugins/props/util/ElementProp.js
LeaVerou: Did we do that before?

DmitrySharabin: We did it in `get`, so, yes: https://github.com/nudeui/element/blob/bb5c2b59740351ec60c1f9a160c0cb6f3b31ed64/src/plugins/props/util/ElementProp.js#L92-104.

DmitrySharabin:
While testing the changes with the color elements, I found a regression. The default value of the color-inline component is calculated based on the component `textContent`:

```js
value: {
	type: String,
	default () {
		return this.textContent.trim();
	},
}
```

The updated architecture where `get` is side-effect-free doesn't support such cases anymore. Here is a regression test.

```js
{
    name: "default() that reads a non-prop (e.g. this.textContent) fires on mount",
    description:
        "A function default fires a mount event when it reads another prop, but not when it reads a non-prop receiver member:" +
        "  default () { return this.base }    // this.base is a prop → fires ✓" +
        "  default () { return this.content } // this.content isn't  → fires ✗ (BUG)" +
        "inferDependencies scrapes the name either way, promoting default() to a synthetic computed prop that depends on `content`. Since no such prop exists, nothing cascades to it and a computed prop doesn't self-fire on mount, so the chain stays silent. The value still derives correctly on read, but the mount propchange never fires.",
    arg: {
        mixin (Class) {
            Object.defineProperty(Class.prototype, "content", {
                get () {
                    return "hi";
                },
            });
        },
        props: {
            value: {
                default () {
                    return this.content;
                },
            },
        },
    },
    expect: [
        ["defaultValue", "hi"],
        ["value", "hi"],
    ],
}
```

LeaVerou:
You mean it doesn't run at all? Or it doesn't re-run?

DmitrySharabin:
> You mean it doesn't run at all? Or it doesn't re-run?

Doesn't run at all. See
image

src/plugins/props/util/PropChangeEvent.js
```diff
@@ -1,29 +1,77 @@
...
+				target.removeAttribute(this.attributeName);
 			}
 			else {
-				target.setAttribute(attributeName, attributeValue);
+				target.setAttribute(this.attributeName, this.attributeValue);
 			}
 		}
-		else if (source === "property") {
-			target[this.name] = value;
+		else {
```
DmitrySharabin: We should still have a guard here, no? Otherwise, we'll break cascade/default tracking.

```suggestion
		else if (this.source) {
```

LeaVerou: Ah ok, I wasn't sure whether we wanted to propagate these or not

LeaVerou:
Pushed a fix for the regression, feel free to add any tests you feel are useful

LeaVerou:
@DmitrySharabin pushed `ca48136` with the fix + your example as a regression test (full attribution in the test description).

Root cause was as you intuited — `get()` being side-effect-free in the refactored model exposed a latent gap. `inferDependencies` promotes any `default() { return this.X }` to a synthetic computed depending on `X`, but when `X` isn't a declared prop (`textContent`, a mixin getter, an instance field, etc.) nothing reactive sits on the other end of that edge, so no cascade ever fires. The pre-refactor code masked this because lazy reads of an orphaned defaultProp chain fired the mount event as a byproduct of `get()` resolving the value.

Fix: every prop's constructor now eagerly derives and fires its own mount event. Single-firing is preserved because a subsequent cascade-driven `update()` short-circuits on the equality check — the eager fire and the cascade can't both emit. Matches the principle that dependencies govern *re*-computation, not whether the initial derivation happens.

src/plugins/props/util/ElementProp.js
DmitrySharabin: With the regression fix in place, this one becomes a real (not hypothetical) issue. If a color element has an invalid value that was set before the component mount/first read, `#derive()` will throw. For example, the invalid `foobar` color in `<color-inline>foobar</color-inline>` can't be parsed on mount, and it throws.

DmitrySharabin (changes requested):
We should fix this regression first: https://github.com/nudeui/element/pull/129#discussion_r3322943309.

I also added [these](https://github.com/nudeui/element/pull/129#issuecomment-4565987850) tests, so please pull.

src/plugins/events/propchange.js
```diff
@@ -18,42 +18,55 @@ const hooks = {
...
-		}
+			for (let aliasName of aliasNames) {
+				this.dispatchEvent(new PropChangeEvent(aliasName, {
+					name: event.name,
+					prop: event.prop,
+					source: event.source,
+					value: event.value,
+					oldValue: event.oldValue,
+				}));
+			}
```
DmitrySharabin, line 68: **Should mount-drain `propchange`s be re-dispatched as semantic alias events?**

This re-broadcasts *every* `propchange` — including the mount-time default fire — as its alias (`spacechange`, `valuechange`, …). So a consumer listening for the alias gets a "change" for a value that never changed (the initial default).

Real case that bit me: a parent listens to a nested child's `spacechange` and writes the child's value back into its own reflected prop:

```js
// child: events = { spacechange: { propchange: "selectedSpace" } }   // selectedSpace is a computed (get) prop
spacePicker.addEventListener("spacechange", () => {
  this.value = spacePicker.value + "." + channelSelect.value; // ← also runs at mount
});
```

At mount the child's computed `selectedSpace` fires with its default (`a98rgb`), this listener re-fires it as `spacechange`, and the parent's authored `value="oklab.a"` gets overwritten (and reflected back to the attribute, destroying the markup). Because it races the parent's own value-sync, it's intermittent.

The catch is that for a **computed** aliased prop you can't filter it by `source`: both the mount fire *and* a genuine post-mount cascade arrive with `source: undefined` (`ElementProp.update()` clears it) — only `oldValue === undefined` separates them. And `PropChangeEvent.applyTo` keys off `source`, so it would also refuse to mirror *real* computed-prop changes.

So: should aliases fire during the mount drain for initial/default values at all — and if they do, what's the intended signal to tell "initial" from "changed"? (Pre-#106 these carried `source: "initial"`, which at least gave consumers a marker to filter on.)

LeaVerou: I think the whole idea of `applyTo()` is fundamentally broken TBH. And for most callsites in color-elements replacing it would be a one line fix, I think. So I wonder if we should just get rid of it…

DmitrySharabin: With the adoption of `updated({change})`, none of the color elements use it. So, yes, we can get rid of it.

src/plugins/props/util/PropChangeEvent.js
```diff
@@ -1,29 +1,77 @@
...
 
-		this.name = name;
-		this.prop = prop;
+		// Assign the rest as own properties, skipping anything Event already
+		// owns (init dict, state like `target`/`defaultPrevented`, methods).
+		for (let [key, value] of Object.entries(options)) {
+			if (!(key in Event.prototype)) {
+				this[key] = value;
+			}
+		}
```
DmitrySharabin, line 52: I wonder if we should handle (ignore) read-only event properties (like `isTrusted`) here to allow the pattern that color elements heavily use: `this.dispatchEvent(new event.constructor(event.type, { ...event }));`. Now, it throws because of `isTrusted`. And it's a regression from the consumer's POV.
There might appear other such properties, and we'll be ready for it.

LeaVerou: Would this fix it?


```suggestion
		for (let [key, value] of Object.entries(options)) {
			if (!(key in Event.prototype) && !(key in this)) {
				this[key] = value;
			}
		}
```

DmitrySharabin: It would. Verified.

DmitrySharabin: How about this alternative? What if instead of walking the prototype chain twice, we just don't catch such cases (that seem legacy in the browser) and leave them no-op?

```js
try {
	this[key] = value;
}
catch {
	// [LegacyUnforgeable] own properties (e.g., isTrusted) are non-writable
}
```

DmitrySharabin:
Submitted a PR with possible fixes if it helps: #132.

DmitrySharabin (approved):

Issue #131, opened 2026-06-01
Guard spec.convert() against throwing


DmitrySharabin:
spec.parse() calls are now guarded with try/catch (#129), but spec.convert() can also throw on values that parsed successfully but aren't valid input for the conversion step.

We should apply the same warn-and-degrade pattern to spec.convert() calls.

Should we degrade/fall back to the parsed value (that is correct on that step)?

LeaVerou:
Yes, but we shouldn't fall back to the parsed value, since it could be a completely different type, we should probably fall back to the previous value

PR #132, opened 2026-06-01
Guard spec.parse(), event cloning, and remove applyTo()

DmitrySharabin:
## Summary
- Wrap `spec.parse()` in `#derive()` with try/catch so invalid pre-mount values (e.g., `<color-inline>foobar</color-inline>`) degrade to `undefined` instead of throwing
- Make the `PropChangeEvent` constructor resilient to `[LegacyUnforgeable]` own properties (e.g., `isTrusted`) so `new event.constructor(event.type, { ...event })` cloning works
- Remove `applyTo()` — fundamentally broken (keys off `source`, refuses real computed-prop changes), and no longer used after `updated({change})`
- `spec.convert()` can also throw but is intentionally left unguarded (out of scope of the original PR since it also fails on `main`) — filed as #131

### DIFF
--- a/src/plugins/props/util/PropChangeEvent.js
+++ b/src/plugins/props/util/PropChangeEvent.js
@@ -45,7 +45,12 @@ export default class PropChangeEvent extends Event {
 		// owns (init dict, state like `target`/`defaultPrevented`, methods).
 		for (let [key, value] of Object.entries(options)) {
 			if (!(key in Event.prototype)) {
-				this[key] = value;
+				try {
+					this[key] = value;
+				}
+				catch {
+					// [LegacyUnforgeable] own properties (e.g. isTrusted) are non-writable
+				}
 			}
 		}
 	}
--- a/src/plugins/props/util/PropChangeEvent.js
+++ b/src/plugins/props/util/PropChangeEvent.js
@@ -44,13 +44,8 @@ export default class PropChangeEvent extends Event {
 		// Assign the rest as own properties, skipping anything Event already
 		// owns (init dict, state like `target`/`defaultPrevented`, methods).
 		for (let [key, value] of Object.entries(options)) {
-			if (!(key in Event.prototype)) {
-				try {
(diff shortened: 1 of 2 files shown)


src/plugins/props/util/PropChangeEvent.js
```diff
@@ -45,33 +45,13 @@ export default class PropChangeEvent extends Event {
 		// owns (init dict, state like `target`/`defaultPrevented`, methods).
 		for (let [key, value] of Object.entries(options)) {
 			if (!(key in Event.prototype)) {
-				this[key] = value;
+				try {
+					this[key] = value;
+				}
+				catch {
+					// [LegacyUnforgeable] own properties (e.g. isTrusted) are non-writable
+				}
```
LeaVerou: I'd rather not use `try .. catch` on a hot path. Why not `key in this`? It would avoid the `isTrusted` issue…

LeaVerou (changes requested):

src/plugins/props/util/PropChangeEvent.js
DmitrySharabin: No strong opinion. It's just one more traversal of the prototype chain, if I'm not mistaken. Right? Good to know that it's better than `try/catch` here. Just curious why (for better understanding).

LeaVerou: `try .. catch` is slower typically, you want to use them for (literal) exceptions, not normal control flow. Search for "Don't Use Exceptions For Flow Control" to read more.

But you're right about the one more traversal, I think we can skip the first check (`key in Event.prototype`)

LeaVerou: Also add a note to NOT use class fields as they'll break this check

DmitrySharabin: Done

LeaVerou (approved):

PR #138, opened 2026-09-18
Let a Boolean prop be set to false

DmitrySharabin:
A `Boolean` prop can never be set to `false`:

```js
static props = { flag: { type: Boolean, default: true } };

el.flag = false;
el.flag;            // true
```

`parse()` is there to collapse an absent attribute — which arrives as `null` — to `false`. But it also runs on property writes, where `false` is a real value rather than an absence, so it parses straight back to `true`:

```js
parse (value) {
	return value !== null;   // parse(false) === true
}
```

The prop then stores `true`, `#derive()` returns `true`, and `equals()` sees no change, so not even a `propchange` fires. The write is silent.

## The fix

Exclude `false` alongside `null`. Attribute semantics are untouched — presence is still truth, including `flag=""` and `flag="false"`.

## Tests

`test/plugins/props/boolean.js`, four cases — one per branch. The first three fail on `main` and pass here:

| | before | after |
| --- | --- | --- |
| `el.flag = false` on a `default: true` prop | `true` | `false` |
| writing `false` to a reflected prop mounted with `flag=""` | attribute stays `""` | removed |
| reactive `default()`: write `false`, then `undefined` | `[true, true]` | `[false, true]` |
| `flag="false"` and `flag=""` both read as `true` | ✅ | ✅ |

The third case is the one that pins the semantics: `false` is a write, `undefined` is an absence, and only the latter re-resolves `default()`.

`npm test`: 196/201 with this branch, and the one remaining failure (`hooks › first_ prefix`) is pre-existing on `main`.

Nothing covers the `value !== null` half, because I could not find a public path that reaches `parse(null)`: `el.flag = null` preserves `null` ahead of `parse`, `removeAttribute` re-resolves the default (or yields `undefined`), and an absent attribute never calls `parse` at all. The guard may be vestigial — left untouched here.

## Why it surfaced

[docspire](https://github.com/docspired/docspire) has a `<foldable-content>` whose `open` state wants to be a reflected Boolean prop so `shadow.css` can select `:host([open])`. It could not be: `open` started `true` for its static type and no write would bring it back down. That element keeps a private field for now.


LeaVerou (approved):
The fact that attributes and properties go through the same path is not ideal, but oh well

